PR #16, opened 2018-12-06
Genia parser

danielkingai2:
Parsing scores on GENIA test: UAS: 0.8944619583117678, LAS: 0.8763939208582319

I think this parser is no longer as good at normal text. A more sophisticated method of training could likely fix this problem (e.g. intersperse examples of normal text with the GENIA text when training the parser). This parser fails some of the sentence boundary detection tests from spacy, some of which were easily fixed and some of which were not. The tests that the GENIA parser newly fails are marked as such with a comment.

The main changes are: add a model folder with just the retrained parser, add a model folder with the tokenizer/splitting stuff and the parser, and add retrain_parser.py script to retrain the parser.


scripts/retrain_parser.py
```diff
@@ -0,0 +1,150 @@
...
+from tqdm import tqdm
+import random
+from spacy.util import minibatch, compounding
+from spacy import util
+from timeit import default_timer as timer
+from spacy.cli.train import print_progress
+import argparse
+import json
+
+def train_parser(train_json_path: str,
```
DeNeutoy: Add a test for this function? It would be relatively straightforward to check that it would run on a tiny test fixture or something.

scripts/retrain_parser.py
```diff
@@ -0,0 +1,150 @@
...
+    meta["author"] = "Allen Institute for Artificial Intelligence"
+    meta["url"] = "allenai.org"
+    meta["sources"] = ["OntoNotes 5", "Common Crawl", "GENIA 1.0"]
+    meta["version"] = "1.0.0"
+    meta["spacy_version"] = ">=2.0.18"
+    meta["parent_package"] = "spacy"
+    meta["speed"] = {"gpu": None, "nwords": nwords, "cpu": cpu_wps}
+    meta["email"] = "ai2-info@allenai.org"
+
+    save final model and output results on the test set
```
DeNeutoy: How does this pass CI? I think this should be a comment.

scripts/retrain_parser.py
```diff
@@ -0,0 +1,150 @@
+import os
+import spacy
+
+from spacy.tokens import Doc
+from spacy.gold import GoldParse, GoldCorpus
+from conllu.parser import parse_line, DEFAULT_FIELDS
```
DeNeutoy: Looks like you've got a bunch of unused imports here.

DeNeutoy:
Oh, having seen the script in the history which you used to convert to the spacy format, I still think this is sub-optimal - the documents should be the actual documents from the abstracts, not just 10 random sentences concatenated together.

scripts/retrain_parser.py
danielkingai2: Ah, just figured out why. (1) there is not test that calls this script (2) pylint is only run on the main module folder (while this is in the scripts folder)

danielkingai2:
@DeNeutoy agree that the difference is likely due to not having pretrained vectors. and agree that it would be good for the models to be in s3 rather than the repo.

As for stitching the abstracts together, I had most of this code written and then realized i could use the spacy convert command instead if i didn't stitch the abstracts together. let me see if i saved that code somewhere...

PR #18, opened 2018-12-10
Med mentions

DeNeutoy:
This adds a script which trains a model on the entity mentions from the [MedMentions](https://github.com/chanzuckerberg/MedMentions) dataset (for entity linking). The F1 is around 60 ish if I use the subset of the data which is cleaner (typically used for Information Retrieval research) and constrained to a smaller number of UMLS semantic types. 

I don't actually have any idea of the overall plan for SciSpacy so maybe this isn't super useful (the dataset format is used by some other biomedical NER datasets though, so we could reuse some of this code regardless). I mainly did this because I wanted to play around with the dataset, as it's quite new and looks quite good.

### DIFF
--- a/scripts/train_ner.py
+++ b/scripts/train_ner.py
@@ -119,13 +119,13 @@ def evaluate(nlp, eval_data):
         # parse dev data with trained model
         doc = nlp(text)
         predicted_spans = [(ent.start_char, ent.end_char, ent.label_) for ent in doc.ents]
-        scorer(predicted_spans, gold_spans)
+        scorer(predicted_spans, gold_spans["entities"])
 
         if i % 1000 == 0 and i > 0:
-            for name, metric in scorer.get_metric():
-                print(f"{name}: \t\t {metric}")
+            for name, metric in scorer.get_metric().items():
+                print(f"{name}: {metric}")
 
-    for name, metric in scorer.get_metric():
+    for name, metric in scorer.get_metric().items():
         print(f"{name}: \t\t {metric}")
 
 plac.call(main, sys.argv[1:])
--- a/scripts/train_ner.py
+++ b/scripts/train_ner.py
@@ -11,6 +11,7 @@ from pathlib import Path
 sys.path.insert(0, os.path.dirname(os.path.abspath(os.path.join(__file__, os.pardir))))
 
 from SciSpaCy.data_util import read_med_mentions
+from SciSpaCy.per_class_scorer import PerClassScorer
 
 @plac.annotations(
         output_dir=("output directory to store model in", "positional", None, str),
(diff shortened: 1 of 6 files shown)


scripts/train_ner.py
```diff
@@ -0,0 +1,146 @@
...
+                print("Saved model to", output_dir_path)
+
+                # test the saved model
+                print("Loading from", output_dir_path)
+                nlp2 = spacy.load(output_dir_path)
+
+            evaluate(nlp, dev_data)
+
+
+def evaluate(nlp, eval_data):
```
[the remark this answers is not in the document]
DeNeutoy: Thanks I didn't know about that - I actually implemented a different scorer because I wanted to see F1 metrics per class.

DeNeutoy:
@danielkingai2 This is actually ready for review - I have some code which reads in the Semantic Type tree, which will mean I can flatten the types down to something more managable (e.g we could try only having a hierarchy of depth X, which would massively reduce the number of types and maybe make the numbers a bit better). I'll add that separately though because it's basically separate.

scripts/train_ner.py
```diff
@@ -0,0 +1,131 @@
...
+            with tqdm.tqdm(total=total, leave=True) as pbar:
+                for batch in minibatch(train_data, size=batch_size):
+                    docs, golds = zip(*batch)
+                    nlp.update(docs, golds, sgd=optimizer, losses=losses, drop=0.01)
+                    pbar.update(len(batch))
+                    if count % 100 == 0 and count > 0:
+                        print('sum loss: %s' % losses['ner'])
+                    count += 1
+
+            # save model to output directory
```
danielkingai2: not sure if you omitted these things intentionally, but spacy makes some suggestions for training their models, namely compounding batch size, decaying dropout, and averaging parameters: https://spacy.io/usage/training#tips.

scripts/train_ner.py
```diff
@@ -0,0 +1,131 @@
...
+    evaluate(nlp, test_data)
+
+def train(model, train_data, dev_data, output_dir, batch_size, n_iter):
+    """Load the model, set up the pipeline and train the entity recognizer."""
+    if model is not None:
+        nlp = spacy.load(model)  # load existing spaCy model
+        print("Loaded model '%s'" % model)
+    else:
+        nlp = spacy.blank('en')  # create blank Language class
+        print("Created blank 'en' model")
```
danielkingai2: you may need to load a specific vocab for this to work with other components

scripts/train_ner.py
DeNeutoy: Oh nice thanks 👍

scripts/train_ner.py
DeNeutoy: For the `vocab_path` in the parser retraining script what do you pass in? I might leave this because eventually all these retraining scripts will need to take a `model_path` parameter which will have been generated from `spacy init-model` for some biomed specific vocab anyway.

danielkingai2: yeah, that is reasonable. I currently pass in the vocab from combined_all_model and then replace the vocab in combined_all_model with the vocab from the retrained model. there is definitely a better way to handle that, just my hacky solution for now

PR #20, opened 2018-12-13
Retrain tagger with parser

danielkingai2:
The results dropped a bit when I made each abstract its own doc (rather than just taking 10 sentences per doc). Training the tagger with the parser doesn't seem to have helped the parser either. The current results are:

UAS: 88.20111767220226
LAS: 83.52462519986702
Tag %: 93.92745820345706

compared to the results from en_core_web_sm on genia:

UAS: 38.79313230027779
LAS: 33.25077813849192
Tag %: 75.40096344573534

and the retrained models results on ontonotes test:

UAS: 48.23223139414712
LAS: 35.89068693502179
Tag %: 77.11257347126586

compared to en_core_web_sm results on ontonotes test:

UAS: 91.66278968907955
LAS: 89.78869815035314
Tag %: 97.40151035187591


danielkingai2:
@DeNeutoy still investigating something wrong with training the tagger...

danielkingai2:
I just copied over your dump to spacy code, which works great! It turns out the thing making the results worse was that iw as starting from a blank en model instead of loading the language class. the new results are:

Retrained genia evaluation
Test results:
UAS: 89.45786382624513
LAS: 87.83931681901562
Tag %: 98.67606412382531
Token acc: 100.0

Retrained ontonotes evaluation
Test results:
UAS: 46.438321905674925
LAS: 35.115040743596694
Tag %: 76.52387845043262
Token acc: 99.95606651991342

DeNeutoy:
Sweet, nice one. FYI we should remember to explore the `gold_preproc=True/False` option, because although it looks like when you are training it the performance is way worse (because if you tokenise stuff wrong it hurts performance), the noise injection thing might really help the sentence splitting. I trained a model with this option just to see what happened and with gold preprocessing at eval time, it got 88.6 UAS and 86.8 LAS, but it's possible it's worth it if the sentence segmentation is much more robust. Do you have examples of hard sentence segmentation from papers etc?

PR #22, opened 2018-12-14
add dump_to_spacy script

DeNeutoy:
@danielkingai2 I think this script does the dumping to file in the right way for this version of spacy - I see what you mean about the differences between this version and later, that will be annoying to change

Edit: I tried converting the data and training a parser with `spacy train en -N train.json dev.json`, this seems to work nicely - 98.6 POS tag acc and 88.8 LAS after 3 epochs of training.

### DIFF
--- a/scripts/dump_to_spacy.py
+++ b/scripts/dump_to_spacy.py
@@ -1,0 +1,91 @@
+
+import os
+import sys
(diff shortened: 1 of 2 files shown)


DeNeutoy:
closed in favour of #20

PR #23, opened 2018-12-14
add makefile shell to fill in

DeNeutoy:
Add a Makefile shell to generate the various components. Eventually I want all of these commands to run without having to download any data, by using the `cached_path` function I added and storing all the data we need in an S3 bucket.

### DIFF
--- a/Makefile
+++ b/Makefile
@@ -1,0 +1,22 @@
+
+
+BUILD_DIR="./build"
+
+init:
+	# Run python script to build vocabulary with vectors.
+	# Adds custom tokeniser to base model.
+	echo "Not implemented"
+
+parser:
+	# Takes in a model output by init and adds "tagger" and "parser" pipelines.
+	echo "Not implemented"
+
+ner:
+	# Takes in a model output by init or parser and adds an "ner" pipeline.
+	echo "Not implemented"
+
+package:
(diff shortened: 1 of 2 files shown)


Makefile
```diff
@@ -0,0 +1,22 @@
...
+	# Takes in a model output by init and adds "tagger" and "parser" pipelines.
+	echo "Not implemented"
+
+ner:
+	# Takes in a model output by init or parser and adds an "ner" pipeline.
+	echo "Not implemented"
+
+package:
+	# Create model packages for 1) The library and 2) The Spacy model.
+	bash scripts/create_model_package.sh ${BUILD_DIR}
```
danielkingai2: should this include the setup.py step and the pip install steps?

DeNeutoy: No - I think we can assume that this will be run inside the docker image, or in an already setup environment. I could add a `make install` step?

danielkingai2: yeah i think having a make install step would be reasonable, just given that its a three step process. unless the intent is to make this an actual python package (ie put it on pypi)

DeNeutoy: I added one and replaced what was in the Dockerfile with `make install` 👍  I think putting it on pip is not so hard, also.

danielkingai2: i think probably not, but downloading the models might be a barrier, not sure how much of a barrier though

PR #24, opened 2018-12-14
Ontonotes mix

danielkingai2:
Simply mixing in some of the ontonoes training data (i used 0.01 as the percentage) doesn't really change performance on genia, and improves performance on ontonotes significanly. certainly fancier things could be done here, but i am going back to mainly working on the competition that i am participating in now. here are the results of the newly trained parser and tagger

Retrained genia evaluation
UAS: 89.38557819773715
LAS: 87.86570891266146
Tag %: 98.64842454394693

Retrained ontonotes evaluation
UAS: 71.47788282743629
LAS: 65.09388995102533
Tag %: 90.93788874450267


scripts/retrain_parser_and_tagger.py
```diff
@@ -13,13 +13,15 @@
...
+import itertools
 
 def train_parser_and_tagger(train_json_path: str,
                             dev_json_path: str,
                             test_json_path: str,
                             vocab_path: str,
                             model_output_dir: str,
-                            ontonotes_path: str = None):
+                            ontonotes_path: str = None,
+                            ontonotes_train_pct: float = 0.0):
```
DeNeutoy: `ontonotes_train_percent` 🙏

PR #26, opened 2018-12-14
Add some typing

danielkingai2:
Please let me know if something looks weird, as I haven't really used typing in python much before


scripts/retrain_parser_and_tagger.py
```diff
@@ -167,13 +167,13 @@ def train_parser_and_tagger(train_json_path: str,
...
-         model_output_dir,
-         ontonotes_path,
-         ontonotes_train_percent):
+def main(train_json_path: str,
+         dev_json_path: str,
+         test_json_path: str,
+         vocab_path: str,
+         model_output_dir: str,
+         ontonotes_path: str,
+         ontonotes_train_percent: str):
```
DeNeutoy: This is a `float`

SciSpaCy/custom_sentence_segmenter.py
```diff
@@ -1,19 +1,21 @@
...
-def combined_rule_sentence_segmenter(doc):
+def combined_rule_sentence_segmenter(doc: Doc) -> Doc:
     """Adds sentence boundaries to a Doc. Intended to be used as a pipe in a spaCy pipeline.
 
     @param doc: the spaCy document to be annotated with sentence boundaries
     """
 
     # keep track of the two previous tokens
-    prev_tokens = [None, None]
+    prev_tokens: List[Any] = [None, None]
```
DeNeutoy: Try not to use `Any`. I think these should be `spacy.Token`? It's fine to declare their type in the type hint but then leave it as `None` in the assignment, you're basically just saying "anything I put in here _will_ have this type"

scripts/spacy_convert.py
```diff
@@ -1,3 +1,5 @@
+from typing impost List, Tuple
```
DeNeutoy: \s\impost\import

scripts/retrain_parser_and_tagger.py
danielkingai2: so it is

SciSpaCy/custom_sentence_segmenter.py
danielkingai2: So they should be tokens but then the type checker fails because they are assigned to None, and i try to access the .text attribute of them and it says None has no attribute text. Is there another way around this?

danielkingai2: I tried Optional[Any] because it seemed like that was the equivalent of a nullable type, but it didn't like that either

DeNeutoy: I think if you separate them it will work, e.g `token1: Token = None, token2: Token = None`

danielkingai2: Thanks, that seems to have worked

PR #27, opened 2018-12-15
add untyped span metrics

DeNeutoy:


SciSpaCy/per_class_scorer.py
```diff
@@ -52,9 +63,12 @@ def get_metric(self, reset: bool = False):
             all_metrics[f1_key] = f1_measure
 
         # Compute the precision, recall and f1 for all spans jointly.
-        precision, recall, f1_measure = self._compute_metrics(sum(self._true_positives.values()),
-                                                              sum(self._false_positives.values()),
-                                                              sum(self._false_negatives.values()))
+        untyped_true_positives = sum({v for k, v  in self._true_positives.items() if k != "untyped"})
```
danielkingai2: If I am reading this correctly, these aren't actually 'untyped_true_positives' etc, but are actually 'sum_of_all_classes_true_positives' or something like that? given that they are used to get the overall metrics, not the untyped metrics.

PR #29, opened 2018-12-17
Lowercase

DeNeutoy:
`SciSpaCy` -> `scispacy`


DeNeutoy:
@danielkingai2 do you have an idea why this Docker image can't build in CI? I can build it locally ...

danielkingai2:
@DeNeutoy do the folders not need to be renamed for this to work?

Dockerfile
```diff
@@ -22,7 +22,7 @@ COPY requirements.in .
 # add the code as the final step so that when we modify the code
 # we don't bust the cached layers holding the dependencies and
 # system packages.
-COPY scispacy/ scispacy
+COPY scispacy/ scispacy/
```
danielkingai2: I think maybe `COPY SciSpaCy/ scispacy/` would work? or renaming the github repo and SciSpaCy folder. I'm not sure why it would work locally but not in teamcity though

DeNeutoy: But this is copying the python package inside the git repo, not the actual repo right? Or is the build step run such that it copies the entire repo?

DeNeutoy: Ok something weird is happening, it didn't pick up my changes to the package...

DeNeutoy:
@danielkingai2 the install works now 👍

danielkingai2:
@kyleclo @acohan we are lowercasing scispacy and the installation instructions are going to change slightly as result. You should just need to change the SciSpaCy in the model path to scispacy

DeNeutoy:
Ya I did that already: https://github.com/allenai/SciSpaCy/pull/29/files#diff-04c6e90faac2675aa89e2176d2eec7d8

danielkingai2:
@DeNeutoy ah my bad, was looking at the wrong version

PR #37, opened 2019-01-15
Genia tokenizer

DeNeutoy:
#### Tokenising 10k pubmed abstracts:
Speed test: genia_tokenizer
iteration 0
iteration 1
iteration 2
Mean (3 runs) 28.98577332496643, Std: 0.9618087746695202.

Speed test: custom_tokenizer
iteration 0
iteration 1
iteration 2
Mean (3 runs) 104.88270195325215, Std: 0.5736707635762457.

Speed test: spacy
iteration 0
iteration 1
iteration 2
Mean (3 runs) 64.94790593783061, Std: 0.6912257483692967.

### DIFF
--- a/scispacy/genia_tokenizer.py
+++ b/scispacy/genia_tokenizer.py
@@ -25,7 +25,6 @@ from spacy.tokens import Doc
 
 INPUT_ENCODING = "UTF-8"
 OUTPUT_ENCODING = "UTF-8"
-DEBUG_GTB_TOKENIZATION = False
 
 # Penn treebank bracket escapes (others excluded)
 PTB_ESCAPES = [('(', '-LRB-'),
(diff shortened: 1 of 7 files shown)


DeNeutoy:
I wanted to get several eyes on this because it's complicated (perhaps unnecessarily). However, it is very fast - 2x faster than the Spacy tokenizer which is written in Cython. I'm pretty pleased with that!

brendan-ai2:
Could you elaborate on how the speed tests were performed? If it was just running via `pytest` by hand, I'd be worried about startup costs dominating.

danielkingai2:
I think it would be good to at least know how many of the original spacy tests this tokenizer fails. I imagine that the spacy tokenizer's long set of special strings/patterns that it handles are what makes it slower, although not sure about that.

scispacy/genia_tokenizer.py
```diff
@@ -0,0 +1,355 @@
...
+# (IL)-mediated", and "p65(RelA)/p50", treat following dashes and
+# slashes as space.  Nested brackets are resolved inside-out;
+# to get this right, add a heuristic considering boundary
+# brackets as "space".
+
+# (First a special case (rareish): "protect" cases with dashes after
+# paranthesized expressions that cannot be abbreviations to avoid
+# breaking up e.g. "(+)-pentazocine". Here, "cannot be abbreviations"
+# is taken as "contains no uppercase charater".)
+_INITIAL.append((re.compile(r'\(([^ ()\[\]{}]+)\)-'), r'-LRB-\1-RRB--'))
```
danielkingai2: I think the comment doesn't match the code here anymore. There is nothing in this regex that excludes uppercase characters.

DeNeutoy:
Hmm I think this fails too many of the spacy tests to be useful. I should have done that earlier...

danielkingai2:
Yeah, sorry it didn't occur to me to suggest that right away

PR #38, opened 2019-01-17
Custom init model

DeNeutoy:
Adds a custom init model script which creates a vocab and optionally vectors. It also adds a bit of functionality:
- Allows not adding all the words in the vectors file to the vocab (this makes the model massive because the pubmed vectors are very large)
- Allows overriding items in the `meta.json` file so that we can have correct info in our models
- Makes providing the vectors actually optional.

### DIFF
--- a/Makefile
+++ b/Makefile
@@ -16,7 +16,7 @@ init-large:
 	en ./build/base_large/ \
 	${PUBMED_FREQS} \
 	-v ${PUBMED_VECTORS} \
-	-x -V 40000 -m ./data/meta_small.json
+	-x -V 40000 -m ./data/meta_large.json
 
 parser:
 	# Takes in a model output by init and adds "tagger" and "parser" pipelines.
--- a/Makefile
+++ b/Makefile
@@ -18,11 +18,6 @@ init-large:
 	-v ${PUBMED_VECTORS} \
 	-x -V 40000 -m ./data/meta_small.json
 
-init:
-	# Run python script to build vocabulary with vectors.
-	# Adds custom tokeniser to base model.
-	echo "Not implemented"
-	python scripts/init_model.py en ./test_init_model ./pubmed.freqs -m ./meta_small.json
 parser:
 	# Takes in a model output by init and adds "tagger" and "parser" pipelines.
 	echo "Not implemented"
@@ -35,7 +30,7 @@ package:
 	# Create model packages for 1) The library and 2) The Spacy model.
 	bash scripts/create_model_package.sh ${BUILD_DIR}
 
-all: init parser ner package
+all-small: init-small parser ner package
 
 install:
 	pip install -r requirements.in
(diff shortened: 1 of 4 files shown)


scripts/init_model.py
```diff
@@ -0,0 +1,174 @@
...
+def init_model(lang, output_dir, freqs_loc=None,
+               vectors_loc=None, no_expand_vectors=False,
+               meta_overrides=None, prune_vectors=-1):
+    """
+    Create a new model from raw data, like word frequencies, Brown clusters
+    and word vectors.
+    """
+    if vectors_loc is not None:
+        vectors_loc = cached_path(vectors_loc)
+    freqs_loc = cached_path(freqs_loc)
```
danielkingai2: freqs_loc is optional right? Should this be inside a None check?

scripts/init_model.py
```diff
@@ -0,0 +1,174 @@
...
+
+    if freqs_loc is not None and not freqs_loc.exists():
+        prints(freqs_loc, title=Messages.M037, exits=1)
+    vectors_loc = ensure_path(vectors_loc)
```
danielkingai2: does this line also need a none check? (I didn't go look at the spacy code for this utility, so maybe its fine)

scripts/init_model.py
```diff
@@ -0,0 +1,174 @@
...
+        return (line.decode('utf8') for line in gzip.open(str(loc), 'r'))
+    elif loc.parts[-1].endswith('zip'):
+        zip_file = zipfile.ZipFile(str(loc))
+        names = zip_file.namelist()
+        file_ = zip_file.open(names[0])
+        return (line.decode('utf8') for line in file_)
+    else:
+        return loc.open('r', encoding='utf8')
+
+def create_model(lang, probs, oov_prob, vectors_data, vector_keys, expand_vectors, prune_vectors):
```
danielkingai2: If I'm reading this correctly, this means that if this function is called with vectors/probs empty/None, as it could be based on optional arguments to init_model, that lang class returned won't have a vocab. Should this function default to using the default vocab?

scripts/init_model.py
DeNeutoy: Done thanks

scripts/init_model.py
DeNeutoy: The plac annotations don't actually specify default values for positional arguments (i.e it crashes) I agree this is very confusing. I was mostly trying to stay close to the spacy script as i'm going to submit these additions upstream, so hopefully eventually we can use them and remove this script

Makefile
```diff
@@ -2,11 +2,27 @@
...
+	en ./build/base_small/ \
+	${PUBMED_FREQS} \
+	-m ./data/meta_small.json
+
+init-large:
+	python scripts/init_model.py \
+	en ./build/base_large/ \
+	${PUBMED_FREQS} \
+	-v ${PUBMED_VECTORS} \
+	-x -V 40000 -m ./data/meta_small.json
```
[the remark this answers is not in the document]
DeNeutoy: Yep, thanks

PR #43, opened 2019-01-22
Build models

DeNeutoy:
Fixes some stuff from actually trying to train the pipeline:

- Be more careful with the ` with nlp.disable_pipes` because this actually removes the pipes from the model when they are under this scope, meaning some parts of the pipeline don't get serialized
- Fixes embarrassing bug I introduced in the `PerClassScorer` which modified the input
- Adds more words to the large model's vocab, in correlation with Spacy

### DIFF
--- a/Makefile
+++ b/Makefile
@@ -51,7 +51,7 @@ ner-small-from-parser:
 	python scripts/train_ner.py \
 	--model_output_dir ${SMALL_NER} \
 	--data_path ${MED_MENTIONS} \
-	--model_path ${SMALL_PARSER} \
+	--model_path ${SMALL_PARSER}/best/ \
 	--iterations 7 \
 	--label_granularity 7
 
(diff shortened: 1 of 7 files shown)


Makefile
```diff
@@ -51,23 +51,24 @@ ner-small-from-parser:
...
+	--model_path ${LARGE_PARSER}/best/ \
 	--iterations 7 \
 	--label_granularity 7
 
 package:
 	# Create model packages for 1) The library and 2) The Spacy model.
 	bash scripts/create_model_package.sh ${BUILD_DIR}
 
-all-small: init-small parser-small ner-small-from-parser package
+all-small: init-small parser-small ner-small-from-parser
```
danielkingai2: do these commands need to have the package command at the end? or is that still tbd

DeNeutoy: still tbd, I think i'm going to bin the makefile approach, it's too unwieldy so this will change again before we release the models

PR #47, opened 2019-01-23
Remove makefile

DeNeutoy:
- Remove makefile
- Make the pipeline really modular, so you can run it all at once or in pieces
- Fix some packaging things (spacy adds the lang to the name when you package it, so it shouldn't be in the name already)


scripts/train_ner.py
```diff
@@ -106,20 +112,21 @@ def train(model, train_data, dev_data, output_dir, n_iter):
...
-    if not output_dir_path.exists():
-        output_dir_path.mkdir()
-
-    with nlp.use_params(optimizer.averages):
-        nlp.to_disk(output_dir_path)
-        print("Saved model to", output_dir_path)
+    best_model_path = Path(output_dir + "/" + "best")
+    shutil.copytree(os.path.join(output_dir, str(best_epoch)),
+                    best_model_path)
+    #with nlp.use_params(optimizer.averages):
```
danielkingai2: is this supposed to be commented out?

danielkingai2: ah i see, maybe just delete it though

scripts/train_parser_and_tagger.py
```diff
@@ -141,18 +144,25 @@ def train_parser_and_tagger(train_json_path: str,
...
-        final_model_path = os.path.join(model_output_dir, "best")
-        os.makedirs(epoch_model_path, exist_ok=True)
-        nlp.to_disk(final_model_path)
-
-    with open(os.path.join(final_model_path, "meta.json"), "w") as meta_fp:
-        meta_fp.write(json.dumps(meta))
+    final_model_path = os.path.join(model_output_dir, "best")
+    shutil.copytree(os.path.join(model_output_dir, "model" + str(best_epoch)),
+                    final_model_path)
+#    with nlp.use_params(optimizer.averages):
```
danielkingai2: same

PR #48, opened 2019-01-23
flesh out website

DeNeutoy:


docs/index.md
```diff
@@ -2,120 +2,82 @@
 layout: default
 ---
 
-Text can be **bold**, _italic_, or ~~strikethrough~~.
+**SciSpaCy is a Python package containing [SpaCy](https://spacy.io/) models for processing _biomedical_ or _clinical_ text.**
```
danielkingai2: maybe add scientific text as well here?

docs/index.md
```diff
@@ -2,120 +2,82 @@
...
+
+nlp = spacy.load("en_core_sci_sm")
+text = """
+Myeloid derived suppressor cells (MDSC) are immature 
+myeloid cells with immunosuppressive activity. 
+They accumulate in tumor-bearing mice and humans 
+with different types of cancer, including hepatocellular 
+carcinoma (HCC).
+"""
+doc = nlp(text)
```
danielkingai2: you also want to call scispacy.custom_tokenizer.remove_new_lines on the raw text here before passing it on to nlp. This is not strictly necessary, but is expected to improve the results for pdf parsed text. The reason it isn't part of the actual model is that spacy has a strict contract of being lossless, and this function removes some extraneous spaces/hyphens

DeNeutoy: Gotcha - i'll leave that out as it's fairly specific to our use case

PR #56, opened 2019-02-04
Upgrade spacy to 2.1.0

danielkingai2:
Not sure what the proper way to upgrade is, since upgrading to 2.1 requires using the spacy-nightly package


scispacy/version.py
```diff
@@ -1,6 +1,6 @@
 _MAJOR = "0"
 _MINOR = "1"
-_REVISION = "0-unreleased"
```
DeNeutoy: Revert this, this is only for actual releases

scripts/train_parser_and_tagger.py
```diff
@@ -95,18 +98,28 @@ def train_parser_and_tagger(train_json_path: str,
     else:
         optimizer = nlp.begin_training(lambda: train_corpus.train_tuples)
     nlp._optimizer = None
-    print("Itn.  Dep Loss  NER Loss  UAS     NER P.  NER R.  NER F.  Tag %   Token %  CPU WPS  GPU WPS")
-    train_docs = train_corpus.train_docs(nlp, projectivize=True)
+
```
DeNeutoy: Did you check that this is set to "true" internally somehow? (I'm pretty sure all the trees are projective, but still)

danielkingai2: yup, its here https://github.com/explosion/spaCy/blob/a3efa3e8d9ca4c54ad61fa7352a485b79acec04b/spacy/gold.pyx#L192

DeNeutoy: Huh now the constructor argument does nothing, weird

scispacy/version.py
DeNeutoy: The idea is that when we do a release, we upgrade from "0.1.0-unreleased" to "0.1.0", do the release, then push a commit which changes to "0.1.1-unreleased", so that our release has a fixed, specified version and if anyone wants to install from master, their version will be "0.1.1-unreleased" so that they know that it wasn't an officially packaged one.

DeNeutoy:
Oh, this is going to be annoying - any tests which use the checked in models are going to fail, because the models are backward incompatible for this version. Ugh.

DeNeutoy:
I think given that we don't actually _use_ the full model (you're only using the tokenizer one, right?) we can just delete it and the test which uses it.

danielkingai2:
yeah, i was just trying to figure out the best way to handle this. I'm going to see if I can just delete them and the tests that require them

danielkingai2:
There were a bunch of tests that were actually using the whole model, it seemed easier for now to just replace the combined_all_model with the new one, and then figure out what to do with the tests later

DeNeutoy:
sure sounds good - I really want to remove all those models, it's the first thing that i'll do once we have an actual release that we can install in scigraph

danielkingai2:
yeah, agreed

PR #58, opened 2019-02-04
Add sentence segmentation pipe and fix pipeline order

danielkingai2:
@DeNeutoy are you opposed to adding the sentence segmentation pipe back in? I think its useful for real body text, and as a place for other potential ad hoc fixes


DeNeutoy:
I'm just a bit wary of adding things to the models which 90% of people won't need. I guess it's ok as it doesn't really do too much magical stuff.

PR #67, opened 2019-02-12
Export umls json

ibeltagy:
Read the META directory of a UMLS release and output a json file that with a list of concepts. 
Each concept has a canonical name, definition, list of types, and list of aliases. 
Running this script on a UMLS_2017_AA_FULL gives the following statistics: 
```
Number of concepts: 3334533
Number of concepts without canonical name (one of the aliases will be used instead): 10803
Number of concepts with no aliases: 1820697
Number of concepts with 1 aliase: 663401
Number of concepts with > 1 aliase: 850435
Number of concepts with no type: 0
Number of concepts with 1 type: 3053977
Number of concepts with > 1 type: 280556
Number of concepts with no definition: 3139611
Number of concepts with definition from preferred sources: 149029
Number of concepts with definition from other sources: 45893
```


scripts/export_uml_json.py
```diff
@@ -0,0 +1,146 @@
...
+"""
+import sys
+import json
+import plac
+
+# preferred definitions (from S2)
+DEF_SOURCES_PREFERRED = {'NCI_BRIDG', 'NCI_NCI-GLOSS', 'NCI', 'GO', 'MSH', 'NCI_FDA'}
+
+
+@plac.annotations(
```
DeNeutoy: Please could you switch to `argparse`, to be consistent with the rest of the codebase. (`init_model` is an exception because it is copied almost verbatim from the spacy command and will be replaced soon)

scripts/export_uml_json.py
```diff
@@ -0,0 +1,146 @@
...
+        umls_meta_directory=('path of the META directory', 'positional', None, str))
+def main(output_filename, umls_meta_directory):
+
+    concepts_filename = '{}/MRCONSO.RRF'.format(umls_meta_directory)
+    types_filename = '{}/MRSTY.RRF'.format(umls_meta_directory)
+    definitions_filename = '{}/MRDEF.RRF'.format(umls_meta_directory)
+    file_descriptors = '{}/MRFILES.RRF'.format(umls_meta_directory)  # to get column names
+
+    concepts_header = types_header = definitions_header = None
+    with open(file_descriptors) as fin:
```
DeNeutoy: As this is code that reads data from a format no-one is familiar with, the format itself needs a bit more description.

Additionally, please separate this script into 4 functions:

`read_umls_file_descriptors`
`read_umls_concepts`
`read_umls_types`
`read_umls_definitions`

and move them into the library itself (e.g `scispacy/umls_utils.py`).

You could then write a small test and documentation for what each of these functions reads in the docstring of the function itself.

scripts/export_uml_json.py
```diff
@@ -0,0 +1,146 @@
...
+            concept = dict(zip(headers, splits))
+
+            if concept['LAT'] != 'ENG' or concept['SUPPRESS'] != 'N':
+                continue  # Keep English non-suppressed concepts only
+
+            # this condition is copied from S2
+            is_canonical = concept['ISPREF'] == 'Y' and concept['TS'] == 'P' and concept['STT'] == 'PF'
+
+            concept_id = concept['CUI']
+            if concept_id not in concept_details:
```
DeNeutoy: What happens if there is a duplicate `concept_id`? Can that happen? If no, why is this necessary?

scripts/export_uml_json.py
```diff
@@ -0,0 +1,146 @@
...
+                concept_details[concept_id] = {'concept_id': concept_id, 'aliases': list(), 'types': list()}
+            concept_name = concept['STR']
+
+            if not is_canonical or 'canonical_name' in concept_details[concept_id]:
+                # not a canonical name or a canonical name already found
+                concept_details[concept_id]['aliases'].append(concept_name)  # add it as an alisase
```
DeNeutoy: \s\alisase\alias

scripts/export_uml_json.py
```diff
@@ -0,0 +1,146 @@
...
+
+            if concept['LAT'] != 'ENG' or concept['SUPPRESS'] != 'N':
+                continue  # Keep English non-suppressed concepts only
+
+            # this condition is copied from S2
+            is_canonical = concept['ISPREF'] == 'Y' and concept['TS'] == 'P' and concept['STT'] == 'PF'
+
+            concept_id = concept['CUI']
+            if concept_id not in concept_details:
+                concept_details[concept_id] = {'concept_id': concept_id, 'aliases': list(), 'types': list()}
```
DeNeutoy: use `[]` rather than `list()`

scripts/export_uml_json.py
```diff
@@ -0,0 +1,146 @@
...
+                print('Headers for MRSTY.RRF: {}'.format(types_header))
+            elif filename == 'MRDEF.RRF':
+                definitions_header = column_names
+                print('Headers for MRDEF.RRF: {}'.format(definitions_header))
+
+    if concepts_header is None or  types_header is None or definitions_header is None:
+        print('ERROR .. column names are not found')
+        exit()
+
+    concept_details = {}  # dictionary of concept_id -> concept_dictionary
```
DeNeutoy: What is a concept dictionary?

I would suggest using a `NamedTuple` here, e.g:

from typing import NamedTuple
```python
class ConceptInformation(NamedTuple):
	concept_id: str
	aliases: List[str]
	types: List[str]
```

scripts/export_uml_json.py
```diff
@@ -0,0 +1,146 @@
...
+            concept = concept_details.get(definition['CUI'])
+            if concept is None:
+                continue
+
+            if 'definition' not in concept or  \
+                is_from_preferred_source and concept['is_from_preferred_source'] == 'N':
+                concept['definition'] = definition['DEF']
+                concept['is_from_preferred_source'] = 'Y' if is_from_preferred_source else 'N'
+
+    print('Number of concepts: {}'.format(len(concept_details)))
```
DeNeutoy: Compute these statistics once in a single pass through the large dictionary, then print them out. Currently you're iterating over the whole thing 10 times unnecessarily.

As scispacy is using python 3.6, you can also use the modern formatting style for python strings:

```python
f"Here is a comment with a {variable} from the code."
```

scripts/export_uml_json.py
```diff
@@ -0,0 +1,146 @@
...
+    print('Number of concepts with definition from preferred sources: {}'.format(len(
+            [1 for c in concept_details.values()
+             if 'is_from_preferred_source' in c and c['is_from_preferred_source'] == 'Y'])))
+    print('Number of concepts with definition from other sources: {}'.format(len(
+            [1 for c in concept_details.values()
+             if 'is_from_preferred_source' in c and c['is_from_preferred_source'] == 'N'])))
+
+    print('Deleting unused fields and choosing a canonical name from aliases ... ')
+    for concept in concept_details.values():
+        # if a concept doesn't have a canonical name, use the first aliase instead
```
DeNeutoy: \s\aliase\alias

scripts/export_uml_json.py
```diff
@@ -0,0 +1,146 @@
...
+        if 'canonical_name' not in concept:
+            aliases = concept['aliases']
+            concept['canonical_name'] = aliases[0]
+            del aliases[0]
+
+        # deleting `is_from_preferred_source`
+        if 'is_from_preferred_source' in concept:
```
DeNeutoy: Is this source knowledge not useful?

scripts/export_uml_json.py
ibeltagy: done

scripts/export_uml_json.py
ibeltagy: done

scripts/export_uml_json.py
ibeltagy: there are duplicate `concept_id`. With the first one, an entry is added to the dict, and for the following ones, they append names to the aliases. I reorganized this code to make it a bit more readable

scripts/export_uml_json.py
ibeltagy: Probably not.

ibeltagy:
All comments have been addressed, @DeNeutoy. 
Thanks for fixing the CI tests @danielkingai2.

scispacy/umls_utils.py
```diff
@@ -0,0 +1,133 @@
+from typing import List, Dict
```
DeNeutoy: Please add small tests for these functions. You can add test fixtures at `tests/fixtures`.

tests/test_umls_utils.py
```diff
@@ -0,0 +1,36 @@
+# pylint: disable=no-self-use,invalid-name
+import unittest
+
+from scispacy import umls_utils
+
+class TestUtil(unittest.TestCase):
+
+    def test_read_umls_concepts(self):
```
danielkingai2: for these tests, I would probably make the fixture smaller and then assert that you get exactly what you expect

scispacy/umls_utils.py
```diff
@@ -0,0 +1,133 @@
...
+    types_filename = 'MRSTY.RRF'
+    headers = read_umls_file_headers(meta_path, types_filename)
+    with open(f'{meta_path}/{types_filename}') as fin:
+        for line in fin:
+            splits = line.strip().split('|')
+            assert len(headers) == len(splits)
+            concept_type = dict(zip(headers, splits))
+
+            concept = concept_details.get(concept_type['CUI'])
+            if concept is not None:
```
danielkingai2: can you add a comment here saying with this might be None

scispacy/umls_utils.py
```diff
@@ -0,0 +1,133 @@
...
+        for line in fin:
+            splits = line.strip().split('|')
+            assert len(headers) == len(splits)
+            definition = dict(zip(headers, splits))
+
+            if definition['SUPPRESS'] != 'N':
+                continue
+            is_from_preferred_source = definition['SAB'] in DEF_SOURCES_PREFERRED
+            concept = concept_details.get(definition['CUI'])
+            if concept is None:
```
danielkingai2: same as above, when might this be None

PR #72, opened 2019-02-21
baseline linking model

ibeltagy:
The goal is to iterate on this baseline until reaching a decent performance before integrating it with spacy. 
This one has the following performance: 
- MedMentions entities not in UMLS: 0
- MedMentions entities found in UMLS: 71062
- Correct linking: 30.19%
- Wrong linking: 3.93%
- No linking: 65.88%

### DIFF
--- a/scripts/linking.py
+++ b/scripts/linking.py
@@ -1,0 +1,103 @@
+"""
+Linking based on simple text normalization and exact matching
+"""
+
+import json
+import argparse
+from collections import defaultdict
+
+def normalize(text):
+    normalized = " ".join(sorted(text.lower().split(" ")))
+    return normalized
+
+def main(medmentions_path, umls_path):
+
+    # reading entity information from umls
+    entities_dict_by_id = {}
+    entities_dict_by_name = defaultdict(set)
+    with open(umls_path) as f:
+        print('loading entities')
+        entities = json.load(f)
+        print('building indices')
+        for entity in entities:
+            # entity id -> entity object  (UMLS calls them concepts, and MedMentions calls them entities)
+            entities_dict_by_id[entity['concept_id']] = entity
+            # normalized entity name -> entity object  (useful for exact matching)
+            entities_dict_by_name[normalize(entity['canonical_name'])].add(entity['concept_id'])
+    print("entities_dict_by_id", len(entities_dict_by_id))
+    print("entities_dict_by_name", len(entities_dict_by_name))
+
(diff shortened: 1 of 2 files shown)


scripts/linking.py
```diff
@@ -0,0 +1,103 @@
...
+                continue
+
+            if gold_entity_id not in entities_dict_by_id:
+                missing_entity_ids.append((mention, gold_entity_type, gold_entity_id))
+                continue
+
+            found_entity_ids.append((mention, gold_entity_type, gold_entity_id))
+
+            # exact match linking
+            normalized_mention = normalize(mention)
```
danielkingai2: could you break this up a bit? I think 'load_umls()', 'read_pubtator()', and 'link()' functions would be good, but definitely at least want a 'link()' function, then that function can be where iteration on the linking procedure happens

scripts/linking.py
```diff
@@ -0,0 +1,103 @@
...
+        for line in f:
+            if line.strip() == "":
+                continue  # empty line is the start of a new paper
+
+            splits = line.strip().split('\t')
+            if len(splits) == 1:
+                continue  # title or abstract, not used for now
+
+            assert len(splits) == 6
+            paper_id, _, _, mention, gold_entity_type, gold_entity_id = splits
```
kyleclo: sry to be butting in here; noticed this line.  In medmentions dataset, entities can have multiple entity types.  for example:
`27117874	1535	1545	prednisone	T109,T121,T125	C0032952`
so instead of `gold_entity_type` you might want to do another `.split(',')` and store types as a variable-length list

scripts/linking.py
```diff
@@ -0,0 +1,103 @@
+"""
+Linking based on simple text normalization and exact matching
+"""
+
+import json
+import argparse
+from collections import defaultdict
+
+def normalize(text):
+    normalized = " ".join(sorted(text.lower().split(" ")))
```
kyleclo: why is there a `sorted` operation here?  doesnt this break usage for exact match?

scripts/linking.py
ibeltagy: done

scripts/linking.py
ibeltagy: good catch. Switched to using `data_utils` instead

scripts/linking.py
ibeltagy: removed

PR #76, opened 2019-02-21
add dockerfile for pip

DeNeutoy:


Dockerfile.pip
```diff
@@ -0,0 +1,21 @@
...
+
+FROM python:3.6.8-jessie
+
+ENV LC_ALL=C.UTF-8
+ENV LANG=C.UTF-8
+
+WORKDIR /stage/scispacy
+
+RUN pip install scispacy
+RUN pip install https://s3-us-west-2.amazonaws.com/ai2-s2-scispacy/releases/v0.1.0/en_core_sci_sm-0.1.0.tar.gz
```
danielkingai2: this should probably at least a be a variable, and echo what version is being tested or something.

PR #87, opened 2019-03-19
Bump up version and provide support for spaCy 2.1

Bharat123rox:
Bump up version in [requirements file](https://github.com/allenai/scispacy/blob/master/requirements.in)

### DIFF
--- a/requirements.in
+++ b/requirements.in
@@ -1,5 +1,5 @@
 numpy
-spacy==2.0.18
+spacy>=2.0.18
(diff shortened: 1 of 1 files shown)


DeNeutoy:
Hi, sorry this change requires a lot more than just changing the version, because we need to retrain all of the scispacy models. Thanks for the PR, but we need to do some stuff internally before we are ready for this, so we'll do it ourselves. Thanks though!

PR #96, opened 2019-04-09
Reduce memory usage for linking at prediction time

ibeltagy:
Currently requires 5GB
Model files are here: `s3://ai2-s2-scispacy/data/linking_model/`
Run the model with this: `python3 scripts/linking.py --medmentions_path data/medmentions --umls_path data/umls_2017_aa_cat0129.json --k 1,2,10,100,1000 --model_path data/linking_model`

### DIFF
--- a/scripts/linking.py
+++ b/scripts/linking.py
@@ -8,7 +8,7 @@ import os.path
 import datetime
 import numpy as np
 from joblib import dump, load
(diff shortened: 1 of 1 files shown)


ibeltagy:
Training still requires 150GB of memory, but addressing that is a different PR

PR #98, opened 2019-04-18
Deduplicated linking

DeNeutoy:
Moves from including each alias per concept to just having a unique set of words, which each have a mapping back to concepts that they are aliases for. Also adds some code around thresholding NN results by similarity.

### DIFF
--- a/scripts/linking.py
+++ b/scripts/linking.py
@@ -71,6 +71,8 @@ class CandidateGenerator:
         The vectorizer used to encode mentions.
     ann_concept_id_list: List[str]
         A list of strings, mapping the indices used in the ann_index to canonical UMLS ids.
+    mention_to_concept: Dict[str, Set[str]], required.
+        A mapping from aliases to canonical ids that they are aliases of.
 
     """
     def __init__(self,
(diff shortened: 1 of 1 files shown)


scripts/linking.py
```diff
@@ -71,19 +71,24 @@ class CandidateGenerator:
         The vectorizer used to encode mentions.
     ann_concept_id_list: List[str]
```
danielkingai2: nit: change parameter name (and description if needed)

DeNeutoy: Done, thanks, FYI having correct docstrings is not a nit in my book

scripts/linking.py
```diff
@@ -246,15 +235,15 @@ def create_tfidf_ann_index(model_path: str, umls_concept_list: List[str]) -> Non
 def load_tfidf_ann_index(model_path: str):
     # `S` for Search. This controls performance at query time. Maximum recommended value is 2000.
     # It makes the query slow without significant gain in recall.
-    efS = 100
+    efS = 1000
```
ibeltagy: Just curious, why increasing `efS` back again ?

DeNeutoy: Oh sorry, I was just playing around. Once we move it into the main codebase we can be more strict about stuff.

PR #99, opened 2019-04-20
add abbreviation detection and tests

DeNeutoy:
I thought this would be generally useful and also specifically might also help the entity linking efforts, because now we'll be able to expand abbreviations which are mentioned in the text, which it pretty common.

### DIFF
--- a/tests/test_abbreviation_detection.py
+++ b/tests/test_abbreviation_detection.py
@@ -1,0 +1,85 @@
+# pylint: disable=no-self-use,invalid-name
+import unittest
+import spacy
+
+from scispacy.abbreviation import AbbreviationDetector, find_abbreviation, filter_matches
+
+class TestAbbreviationDetector(unittest.TestCase):
+
+
+    def setUp(self):
+        super().setUp()
+
+        self.nlp = spacy.load("en_core_web_sm")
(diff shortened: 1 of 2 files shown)


tests/test_abbreviation_detection.py
```diff
@@ -0,0 +1,93 @@
...
+        # Attribute should be registered.
+        doc = self.nlp(self.text)
+        assert doc._.abbreviations == []
+        doc2 = self.detector(doc)
+        assert len(doc2._.abbreviations) == 2
+
+        long, shorts = doc2._.abbreviations[0]
+        assert long.string == "Spinal and bulbar muscular atrophy "
+        assert len(shorts) == 2
+        assert {x.string for x in shorts} == {"SBMA", "SBMA "}
```
danielkingai2: Should whitespace be stripped from abbreviations?

DeNeutoy: I think this is because `Span.string` includes the whitespace tokens "own", so if you just print `{x for x in shorts}` you wouldn't get that. I think it's probably ok as is.

danielkingai2: Ah, got it. I think I was used to using x.text which doesn't show the whitespace

PR #101, opened 2019-04-25
Add an evaluation option to use a spacy model for ner instead of gold mentions

danielkingai2:
Evaluates candidate generation using a spacy model for ner instead of gold mentions. In this setting, an entity candidate is considered correct if that entity is in the gold entities for a given abstract. i.e. candidate entities are evaluated for correctness at the document level instead of the mention level, to avoid the complexity of trying to align mentions produced by the spacy model with the gold mentions.

### DIFF
--- a/scripts/linking.py
+++ b/scripts/linking.py
@@ -10,7 +10,7 @@ import json
 import argparse
 import datetime
 from collections import defaultdict
-
+from tqdm import tqdm
 
 import scipy
 import numpy as np
@@ -382,7 +382,7 @@ def eval_candidate_generation(examples: List[data_util.MedMentionExample],
             all_golds = []
             all_mentions = []
 
-            for i, example in enumerate(examples):
+            for i, example in tqdm(enumerate(examples), desc="Iterating over examples"):
                 entities = [entity for entity in example.entities if entity.umls_id in umls_concept_dict_by_id]
                 gold_umls_ids = [entity.umls_id for entity in entities]
                 mention_texts = [entity.mention_text for entity in entities]
--- a/scripts/linking.py
+++ b/scripts/linking.py
@@ -365,6 +365,7 @@ def eval_spacy_mentions(examples: List[data_util.MedMentionExample],
             num_candidates = []
             num_filtered_candidates = []
 
+            all_golds_per_doc_set = []
             all_golds = []
             all_mentions = []
             for example, mention_texts, gold_umls_ids in examples_with_labels:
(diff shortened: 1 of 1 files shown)


scripts/linking.py
```diff
@@ -289,37 +297,147 @@ def get_mention_text_and_ids(data: List[data_util.MedMentionExample],
...
-                                                                                   spacy_format=False)
+def eval_spacy_mentions(examples: List[data_util.MedMentionExample],
+                        umls_concept_dict_by_id: Dict[str, Dict],
+                        candidate_generator: CandidateGenerator,
+                        k_list: List[int],
+                        thresholds: List[float],
+                        spacy_model: str):
+    """
+    Evaluates candidate generation using mentions produced by a spacy model. This means that an entity is considered
+    correct if that entity appears anywhere in the abstract
```
ibeltagy: I started working on the supervised model but then realized that we need the mention-level evaluation (not just the document-level). It would be great to add that as well.

ibeltagy: With mention-level evaluation, it would be better to merge this function with `eval_gold_mentions` into one to reduce duplicate code. Or instead,  update `eval_gold_mentions` so that it works with gold or predicted mentions.

scripts/linking.py
```diff
@@ -289,37 +297,147 @@ def get_mention_text_and_ids(data: List[data_util.MedMentionExample],
...
+
+            print(f'MedMentions entities not in UMLS: {len(missing_entity_ids)}')
+            print(f'MedMentions entities found in UMLS: {len(all_golds)}')
+            print(f'K: {k}, Filtered threshold : {threshold}')
+            print('Gold concept in candidates: {0:.2f}%'.format(100 * entity_correct_links_count / len(all_golds_per_doc_set)))
+            print('Gold concepts missed: {0:.2f}%'.format(100 * entity_missed_count / len(all_golds_per_doc_set)))
+            print('Candidate generation failed: {0:.2f}%'.format(100 * mention_no_links_count / len(all_mentions)))
+            print("Mean, std, min, max candidate ids: ", np.mean(num_candidates), np.std(num_candidates), np.min(num_candidates), np.max(num_candidates))
+            print("Mean, std, min, max filtered candidate ids: ", np.mean(num_filtered_candidates), np.std(num_filtered_candidates), np.min(num_filtered_candidates), np.max(num_filtered_candidates))
+
```
ibeltagy: Can you also print the time it took to run this function. I think it will be a good estimate for the run time of the full pipeline

scripts/linking.py
DeNeutoy: Agree with Iz here, you could do this nicely by making `examples` (doc, MedMentionsExample) tuples, which would provide all the information needed to evaluate either way.

scripts/linking.py
```diff
@@ -331,32 +372,144 @@ def main(medmentions_path: str, umls_path: str, model_path: str, ks: str, thresh
...
+                            entity_wrong_links_count += 1
                 else:
-                    entity_wrong_links_count += 1
-                    # print(f'Wrong candidates. Mention Text: {mention_text}, Canonical Name: {gold_canonical_name}')
+                    doc = docs[i]
+                    ner_entities = [ent for ent in doc.ents]
+                    ner_mentions = [ent.text for ent in doc.ents]
+
+                    doc_candidates = set()
+                    doc_golds = set(gold_umls_ids)
```
ibeltagy: redundant

PR #103, opened 2019-05-01
Improving linking evaluation code

ibeltagy:
- fix a small bug in linking evaluation
- refactoring to reduce duplicate code


scripts/linking.py
```diff
@@ -385,72 +384,46 @@ def eval_candidate_generation(examples: List[data_util.MedMentionExample],
...
                         span_from_doc = doc.char_span(gold_entity.start, gold_entity.end)
                         candidates = {}
                         for j, predicted_entity in enumerate(ner_entities):
                             if predicted_entity == span_from_doc:
                                 candidates = batch_candidate_neighbor_ids[j]
                                 break
 
-                        # Keep only canonical entities for which at least one mention has a score less than the threshold.
-                        filtered_ids = {k: v for k, v in candidates.items() if any([z[1] <= threshold for z in v])}
-                        num_candidates.append(len(candidates))
```
ibeltagy: @danielkingai2, I deleted this line and the following two because these counts have already been updated in the loop above. I wanted to double check with you that this is correct

danielkingai2: Ah, yes. that is correct.

PR #106, opened 2019-05-01
Train a linking classifier

ibeltagy:
- Generate training data for the linking classifier
- Train linker


scripts/train_linker.py
```diff
@@ -0,0 +1,81 @@
...
+    for data in lists:
+        features  = [featurizer(d) for d in data]
+        x.append(features)
+        y.append([d['label'] for d in data])
+
+    end_time = datetime.datetime.now()
+    total_time = end_time - start_time
+    print(f'Time: {total_time.total_seconds()} seconds')
+
+    cls = GradientBoostingClassifier(verbose=1)
```
danielkingai2: did you do any hyperparameter search? I think its usually pretty important for gradient boosting trees classifiers

scripts/linking.py
```diff
@@ -11,7 +11,7 @@
 import datetime
 from collections import defaultdict
 from tqdm import tqdm
-
+import jsonlines
```
DeNeutoy: Can we not use 3rd party libraries for this, given you can easily achieve the same thing with `json`.

scripts/linking.py
```diff
@@ -534,8 +542,23 @@ def main(medmentions_path: str,
...
+                                                                    use_gold_mentions, spacy_model, generate_linker_data)
+            with jsonlines.open(filename, 'w') as f:
+                for d in supervised_data:
+                    f.write(d)
+    else:
+        print('Results on the DEV set')
+        eval_candidate_generation_and_linking(dev_examples, umls_concept_dict_by_id, candidate_generator, linking_classifier, k_list, thresholds,
+                                              use_gold_mentions, spacy_model, generate_linker_data)
+
+        print('Results on the TEST set')
```
DeNeutoy: Let's remove this so we don't peek at the test set by default when running the script.

scripts/linking.py
```diff
@@ -358,7 +358,8 @@ def eval_candidate_generation_and_linking(examples: List[data_util.MedMentionExa
                                           k_list: List[int],
                                           thresholds: List[float],
                                           use_gold_mentions: bool,
-                                          spacy_model: str):
+                                          spacy_model: str,
+                                          generate_linking_classifier_training_data: bool):
```
DeNeutoy: If you call this function with multiple k or thresholds, it will duplicate the linking classifier data. Please could you prevent that, preferably by pulling out this functionality into a separate function which only accepts a single k/threshold.

scripts/linking.py
```diff
@@ -463,9 +465,13 @@ def eval_candidate_generation_and_linking(examples: List[data_util.MedMentionExa
                     for candidate_id in candidate_ids:
                         has_definition = 'definition' in umls_concept_dict_by_id[candidate_id]
                         cosine_scores = [cosine for alias, cosine in candidates[candidate_id]]
-                        classifier_example = ({'has_definition': has_definition, 'cosines': cosine_scores,
+                        label = int(gold_entity.umls_id == candidate_id)
```
DeNeutoy: Please can you pull this out into its own function,

`generate_linking_classifier_training_data`. It has some small amount of overlapping code but it would be much cleaner and easier to read.

scripts/train_linker.py
```diff
@@ -0,0 +1,72 @@
+from sklearn.ensemble import GradientBoostingClassifier
+from sklearn.metrics import accuracy_score, classification_report
+from joblib import dump, load
+import jsonlines
```
DeNeutoy: Please don't use jsonlines.

scripts/train_linker.py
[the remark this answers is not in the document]
DeNeutoy: Or... classifier :rage1:

scripts/train_linker.py
```diff
@@ -0,0 +1,72 @@
...
+from tqdm import tqdm
+
+
+def main(data_path: str):
+    train = []
+    dev = []
+    test = []
+    root = ''
+    files = [f'{data_path}/train.jsonl', f'{data_path}/dev.jsonl', f'{data_path}/test.jsonl']
+    lists = [train, dev, test]
```
DeNeutoy: Don't use `lists` as a variable name. It has no meaning. Maybe `data_splits`?

scripts/train_linker.py
```diff
@@ -0,0 +1,72 @@
...
+    # limits = [40000000, 4000000, 4000000]
+    limits = [4000000, 0, 4000000]
+
+    start_time = datetime.datetime.now()
+
+    # load data from files
+    for filename, data, limit in zip(files, lists, limits):
```
DeNeutoy: This whole block took me about 5 minutes to parse, when it should have taken me 30 seconds. 

1. You are looping over the data twice, once to read it and once to featurize it. 
2. It's not really necessary to create all these data lists beforehand, when you are just building a list of them anyway. why not just define `data = []` inside the loop body? It would make the code much easier to read because I don't have to go hunting to find what `data` is.

scripts/train_linker.py
```diff
@@ -0,0 +1,72 @@
...
+        with jsonlines.open(filename) as f:
+            for line in tqdm(f, total=limit):
+                data.append(line)
+                if len(data) >= limit:
+                    break
+    x = []
+    y = []
+    for data in tqdm(lists):
```
DeNeutoy: Don't loop twice over the data

scripts/train_linker.py
```diff
@@ -0,0 +1,72 @@
...
+
+def main(data_path: str):
+    train = []
+    dev = []
+    test = []
+    root = ''
+    files = [f'{data_path}/train.jsonl', f'{data_path}/dev.jsonl', f'{data_path}/test.jsonl']
+    lists = [train, dev, test]
+    # limits = [40000000, 4000000, 4000000]
+    limits = [4000000, 0, 4000000]
```
DeNeutoy: Why are we not reading the dev data, but the test data only?

scripts/linking.py
```diff
@@ -358,7 +358,8 @@ def eval_candidate_generation_and_linking(examples: List[data_util.MedMentionExa
                                           k_list: List[int],
                                           thresholds: List[float],
                                           use_gold_mentions: bool,
-                                          spacy_model: str):
```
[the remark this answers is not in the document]
ibeltagy: done

scripts/train_linker.py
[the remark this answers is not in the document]
ibeltagy: done

ibeltagy: tried changing number of estimators and number of levels, no significant impact

scripts/linking.py
ibeltagy: done

scripts/linking.py
ibeltagy: ok

scripts/linking.py
ibeltagy: added a check

scripts/linking.py
ibeltagy: I doubt the code overlap is going to be small, but the other issue is that with every change in the candidate generation here, we will need to update the other function as well. That's why I think it is better to keep it here. Also, with the code refactoring, it looks a lot better.

scripts/train_linker.py
ibeltagy: done

scripts/train_linker.py
ibeltagy: ok

scripts/train_linker.py
ibeltagy: is this better ?

scripts/train_linker.py
ibeltagy: Just to save time, the dataset is very large compared to the features we have. For validation, sklearn classifiers automatically split the training set into train and dev

PR #107, opened 2019-05-02
Mention eval

danielkingai2:
Makes the mention level evaluation more generous by adding candidates from all spacy spans that have any overlap with the gold span


scripts/linking.py
```diff
@@ -409,11 +409,25 @@ def eval_candidate_generation(examples: List[data_util.MedMentionExample],
...
+                            # one case is that the spacy span has an extra period attached to the end of it
+                            span_from_doc = doc.char_span(gold_entity.start, gold_entity.end+1)
+
                         candidates = {}
-                        for j, predicted_entity in enumerate(ner_entities):
-                            if predicted_entity == span_from_doc:
-                                candidates = batch_candidate_neighbor_ids[j]
-                                break
+                        if span_from_doc is not None:
+                            for j, predicted_entity in enumerate(ner_entities):
```
ibeltagy: are the if statements equivalent to
```i0 = max(span_from_doc.start_char, predicted_entity.start_char); // lower bound of overlap
i1 = min(span_from_doc.end_char, predicted_entity.end_char); // upper bound of overlap
if i0 <= i1:
    overlap
else: 
   no overlap```

danielkingai2: I think so, yeah. I prefer mine though just because they're easier to read and reason about, but am happy to change if you feel strongly.

PR #109, opened 2019-05-05
Run entity-level evaluation on the generated dataset for faster iterations

ibeltagy:


scripts/train_linker.py
```diff
@@ -7,60 +7,103 @@
...
+                            classifier_correct_predictions[k] += 1
+                        classifier_total_predictions[k] += 1
+            candidate_features = []
+            gold_index = -1
+
+        assert candidate_index == len(candidate_features)
+        if d['label'] == 1:
+            assert gold_index == -1
+            gold_index = candidate_index
+        candidate_features.append(all_features[i])
```
DeNeutoy: It's counter-intuitive to me why this is at the end of the loop. Doesn't this mean you miss the first data point? Can you refactor to loop over the data such as:

```
for features, d in zip(all_features, d):
....
```

The reason it's counter-intuitive is because you use `candidate_features` before it is defined in terms of how I read through this code, i.e top to bottom. The more you can avoid things like this, the better.

scripts/train_linker.py
```diff
@@ -7,60 +7,103 @@
...
-def main(data_path: str):
+def entity_level_eval(data, all_features, classifier):
+    k_list = [1, 3, 5, 10]
+    classifier_correct_predictions = defaultdict(int)
+    classifier_total_predictions = defaultdict(int)
+    candidate_features = []
+    gold_index = -1
+    for i, d in tqdm(enumerate(data)):
+        candidate_index = d['candidate_index']
+        if candidate_index == 0:
```
DeNeutoy: Can you collapse these if statements into a single block, so the code is less indented? It makes it easier to read.

Additionally, if you refactor to loop over candidate features from the start, you won't have to check its length here.

scripts/train_linker.py
```diff
@@ -7,60 +7,103 @@
...
+def main(data_path: str, train:bool):
     start_time = datetime.datetime.now()
+    linking_classifier_path = f'{data_path}/linking_classifier.joblib'
 
-    x_train, y_train = read_file(f'{data_path}/train.jsonl', 5000000)  # the full set is unnecessarily large
-    x_dev, y_dev = read_file(f'{data_path}/dev.jsonl', 1)  # the full set is unnecessarily large
-    x_test, y_test = read_file(f'{data_path}/test.jsonl', 5000000)
+    if train:
+        print('Reading training set')
+        d_train, x_train, y_train = read_file(f'{data_path}/train.jsonl', 330000)  # the full set is unnecessarily large
```
DeNeutoy: Does clipping like this introduce any bias? Would it be better to load the whole data, shuffle and then clip?

scripts/train_linker.py
```diff
@@ -7,60 +7,103 @@
...
+                    sorted_ids = np.argsort(-scores[:, 1], kind='mergesort')
+                    for k in k_list:
+                        if gold_index in sorted_ids[:k]:
+                            classifier_correct_predictions[k] += 1
+                        classifier_total_predictions[k] += 1
+            candidate_features = []
+            gold_index = -1
+
+        assert candidate_index == len(candidate_features)
+        if d['label'] == 1:
```
DeNeutoy: Why does the label of the current piece of data effect the processing of the next piece of data? I do not understand what's going on here....

PR #110, opened 2019-05-05
More linker features

ibeltagy:
Added more fields to the linker classifier training data and added more features. None of the additional features help.


ibeltagy:
@DeNeutoy @danielkingai2 , this PR is to share ideas of features. None of the additional features here is helpful

PR #111, opened 2019-05-07
Abbr linker

DeNeutoy:
Changes:

1. A spacy model is now required, because we need the doc to do abbreviation dectection, even with gold mentions.
2. Refactored a bit, pulling out some functions
3. Moved the check for generating the training data outside of the main evaluation function so the script fails faster.
4. Refactored to actually use the doc creation function, (additionally pulls out the generation of the gold and predicted mentions, because I need to be able to modify them when replacing abbreviations).

Both results below use gold mentions. 
**Without abbreviations**
Gold concept in candidates: 83.67%
Gold concept not in candidates: 16.29%
Doc level gold concept in candidates: 92.79%
Doc level gold concepts missed: 7.21%

**With abbreviations**
Gold concept in candidates: 86.60%
Gold concept not in candidates: 13.36%
Doc level gold concept in candidates: 92.77%
Doc level gold concepts missed: 7.23%

### DIFF
--- a/scripts/linking.py
+++ b/scripts/linking.py
@@ -395,6 +395,17 @@ def get_mention_text_and_ids(data: List[data_util.MedMentionExample],
 
     return mention_texts, gold_umls_ids, missing_entity_ids
 
+
+
+def maybe_substitute_span(doc, entity, abbreviations):
+    maybe_ent_span = doc.char_span(entity.start, entity.end)
+    if maybe_ent_span is None:
+        maybe_ent_span = doc.char_span(entity.start, entity.end + 1)
+    if maybe_ent_span in abbreviations:
+        return maybe_ent_span, str(abbreviations[maybe_ent_span])
+    else:
+        return None, None
+
 def get_mention_text_and_ids_by_doc(data: List[data_util.MedMentionExample],
                                     umls: Dict[str, Any],
                                     nlp: Language,
@@ -415,16 +426,6 @@ def get_mention_text_and_ids_by_doc(data: List[data_util.MedMentionExample],
     substituted = 0
     total = 0
 
-
-    def maybe_substitute_span(doc, entity, abbreviations):
-        maybe_ent_span = doc.char_span(entity.start, entity.end)
-        if maybe_ent_span is None:
-            maybe_ent_span = doc.char_span(entity.start, entity.end + 1)
-        if maybe_ent_span in abbreviations:
-            return maybe_ent_span, str(abbreviations[maybe_ent_span])
-        else:
-            return None, entity.mention_text
-
     for example in data:
 
         doc = nlp(example.text)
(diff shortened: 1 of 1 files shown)


scripts/linking.py
```diff
@@ -405,33 +421,94 @@ def get_mention_text_and_ids_by_doc(data: List[data_util.MedMentionExample],
...
+
+        for entity in doc.ents:
+            new_span, _ = maybe_substitute_span(doc, entity, abbreviations)
+            if new_span is None:
+                predicted_mention_texts.append(entity)
+            else:
+                # We have to manually create a new span with the new start and end points, but with the old label,
+                # as spans are read only views of a document.
+                span_with_label = Span(doc, start=new_span.start, end=new_span.end, label=entity.label_)
+                predicted_mention_texts.append(span_with_label)
```
danielkingai2: do you need to also remove the old entity span here?

scripts/linking.py
```diff
@@ -405,33 +421,94 @@ def get_mention_text_and_ids_by_doc(data: List[data_util.MedMentionExample],
...
+
+    return candidates, mention_types
+
+
 def eval_candidate_generation_and_linking(examples: List[data_util.MedMentionExample],
                                           umls_concept_dict_by_id: Dict[str, Dict],
                                           candidate_generator: CandidateGenerator,
-                                          k_list: List[int],
-                                          thresholds: List[float],
+                                          k_list: int,
```
danielkingai2: why did you remove the list types here?

scripts/linking.py
```diff
@@ -448,8 +525,8 @@ def eval_candidate_generation_and_linking(examples: List[data_util.MedMentionExa
     candidate_generator: CandidateGenerator
         A CandidateGenerator instance for generating linking candidates for mentions
     k_list: List[int]
-        A list of k values determining how many candidates are generated
-    thresholds: List[float]
+        A list of values determining how many candidates are generated.
+    thresholds: float
```
danielkingai2: types here are now inconsistent with the parameters

scripts/linking.py
danielkingai2: Looking at this again, I think it doesn't introduce any bugs, but might be a bit confusing to have more entities on the doc than on the MedMentionsExample, but I don't think it matters

scripts/linking.py
DeNeutoy: because I am dumb

scripts/linking.py
DeNeutoy: see above

scripts/linking.py
DeNeutoy: I added a comment that they might not match - it's ok because we don't use the `doc.ents` in the main function anymore (but you are correct that it's important that we don't, because it would mean the abbreviations are not included regardless of the flag.)

PR #112, opened 2019-05-08
Doc level eval

danielkingai2:
```
python scripts/linking.py --medmentions_path s3_data/med_mentions --umls_path s3_data/umls_2017_aa_cat0129.json --model_path data/linking_model2/ --ks 80 --thresholds 1.0 --spacy_model en_core_sci_md --use_soft_matching

K: 80, Filtered threshold : 1.0
Gold concept in candidates: 71.60%
Gold concept not in candidates: 17.35%
Doc level gold concept in candidates: 80.65%
Doc level gold concepts missed: 19.35%
Candidate generation failed: 11.05%
Mention linking precision 46.61%
Doc linking precision 42.67%
Normalized doc linking precision 60.04%
Doc linking recall 48.42%
Linking mention-level recall@1: 41.98%
Normalized linking mention-level recall@1: 58.63%
Linking mention-level recall@3: 54.96%
Normalized linking mention-level recall@3: 76.76%
Linking mention-level recall@5: 59.02%
Normalized linking mention-level recall@5: 82.43%
Linking mention-level recall@10: 63.95%
Normalized linking mention-level recall@10: 89.31%
Mean, std, min, max candidate ids: 56.09, 13.64, 0, 164
Mean, std, min, max filtered candidate ids: 56.09, 13.64, 0, 164
```

### DIFF
--- a/scripts/linking.py
+++ b/scripts/linking.py
@@ -484,12 +484,13 @@ def get_predicted_mention_candidates_and_types(span,
                                                predicted_mention_types,
                                                use_soft_matching):
     """
-    This function returns two lists, candidates, and mention types. These lists will be the same length and have
+    This function returns three lists, candidates, mention types, and mention spans. These lists will be the same length and have
     length equal to the number of predicted entities that overlap with the input gold entity. When not using soft mentions,
     this length will be equal to one, as only one predicted entity can exactly match a gold entity.
     """
     candidates = []
     mention_types = []
+    mention_spans = []
 
     if span is not None:
         for j, predicted_entity in enumerate(ner_entities):
@@ -514,8 +515,9 @@ def get_predicted_mention_candidates_and_types(span,
                 if overlaps:
                     candidates.append(filtered_batch_candidate_neighbor_ids[j])
                     mention_types.append(predicted_mention_types[j])
+                    mention_spans.append(predicted_entity)
 
-    return candidates, mention_types
+    return candidates, mention_types, mention_spans
 
 
 def eval_candidate_generation_and_linking(examples: List[data_util.MedMentionExample],
@@ -584,6 +586,9 @@ def eval_candidate_generation_and_linking(examples: List[data_util.MedMentionExa
             doc_mention_no_links_count = 0  # number of ner mentions that did not have any linking candidates
             doc_num_candidates = []
             doc_num_filtered_candidates = []
+            doc_linking_correct_count = 0
+            doc_linking_golds_in_candidates = 0
+            doc_linking_total_predictions = 0
 
             all_golds_per_doc_set = []
             all_golds = []
@@ -591,8 +596,8 @@ def eval_candidate_generation_and_linking(examples: List[data_util.MedMentionExa
 
             gold_entities_linker_correct = defaultdict(int)
             gold_entities_linker_incorrect = defaultdict(int)
-            predicted_entities_linker_correct = defaultdict(int)
-            predicted_entities_linker_incorrect = defaultdict(int)
+            predicted_entities_linker_correct = defaultdict(set)
+            predicted_entities_linker_incorrect = defaultdict(set)
 
             for doc, example, gold_entities, predicted_entities, gold_umls_ids in tqdm(examples_with_text_and_ids,
                                                                                     desc="Iterating over examples",
@@ -604,6 +609,7 @@ def eval_candidate_generation_and_linking(examples: List[data_util.MedMentionExa
                 doc_golds = set(gold_umls_ids)
                 doc_candidates = set()
                 doc_linker_predictions = set()
+                doc_all_entities_in_candidates = set()
 
                 if use_gold_mentions:
                     mention_texts = gold_entities
@@ -622,8 +628,9 @@ def eval_candidate_generation_and_linking(examples: List[data_util.MedMentionExa
                     num_candidates.append(len(candidate_neighbor_ids))
                     num_filtered_candidates.append(len(filtered_ids))
                     doc_candidates.update(filtered_ids)
-                    mention_types.update(predicted_mention_types[j])
-                    sorted_candidate_ids = linker.link(filtered_ids, mention_text, mention_types)
+                    sorted_candidate_ids = linker.link(filtered_ids, mention_text, predicted_mention_type)
+                    doc_all_entities_in_candidates.update(filtered_ids)
+                    doc_linker_predictions.add(sorted_candidate_ids[1])
 
                 for i, gold_entity in enumerate(entities):
                     if use_gold_mentions:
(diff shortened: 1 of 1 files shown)


scripts/linking.py
```diff
@@ -481,30 +481,43 @@ def get_mention_text_and_ids_by_doc(data: List[data_util.MedMentionExample],
...
-            if overlaps:
-                candidates.update(filtered_batch_candidate_neighbor_ids[j])
-                mention_types.update(predicted_mention_types[j]) 
-
-    return candidates, mention_types
+            if use_soft_matching and span == predicted_entity:
+                candidates = filtered_batch_candidate_neighbor_ids[j]
+                mention_types = predicted_mention_types[j]
+                break
+            elif not use_soft_matching:
```
DeNeutoy: Shouldn't this be `elif use_soft_matching`?

scripts/linking.py
```diff
@@ -569,14 +586,18 @@ def eval_candidate_generation_and_linking(examples: List[data_util.MedMentionExa
...
             all_golds_per_doc_set = []
             all_golds = []
             all_mentions = []
 
-            classifier_correct_predictions = defaultdict(int)
-            classifier_wrong_predictions = defaultdict(int)
-
+            gold_entities_linker_correct = defaultdict(int)
+            gold_entities_linker_incorrect = defaultdict(int)
+            predicted_entities_linker_correct = defaultdict(set)
```
DeNeutoy: This is a little confusing because it's different from whether we are using `gold_mentions` or not, is there a better name? Maybe not...

scripts/linking.py
```diff
@@ -481,30 +481,43 @@ def get_mention_text_and_ids_by_doc(data: List[data_util.MedMentionExample],
...
-                or predicted_entity.start_char <= span.end_char and predicted_entity.end_char >= span.end_char:
-                overlaps = True
-            if overlaps:
-                candidates.update(filtered_batch_candidate_neighbor_ids[j])
-                mention_types.update(predicted_mention_types[j]) 
-
-    return candidates, mention_types
+            if use_soft_matching and span == predicted_entity:
+                candidates = filtered_batch_candidate_neighbor_ids[j]
+                mention_types = predicted_mention_types[j]
```
ibeltagy: do you need to set `mention_spans` as well ?

scripts/linking.py
```diff
@@ -597,13 +620,18 @@ def eval_candidate_generation_and_linking(examples: List[data_util.MedMentionExa
...
+                for candidate_neighbor_ids, mention_text, predicted_entity, predicted_mention_type \
+                    in zip(batch_candidate_neighbor_ids, mention_texts, predicted_entities, predicted_mention_types):
                     # Keep only canonical entities for which at least one mention has a score less than the threshold.
                     filtered_ids = {k: v for k, v in candidate_neighbor_ids.items() if any([z[1] <= threshold for z in v])}
                     filtered_batch_candidate_neighbor_ids.append(filtered_ids)
                     num_candidates.append(len(candidate_neighbor_ids))
                     num_filtered_candidates.append(len(filtered_ids))
                     doc_candidates.update(filtered_ids)
+                    if len(filtered_ids) != 0:
+                        sorted_candidate_ids = linker.link(filtered_ids, mention_text, predicted_mention_type)
```
ibeltagy: why is this being called here ?

scripts/linking.py
ibeltagy: and the one above needs to be `if span == predicted_entity:`

danielkingai2: wow yeah, got that backwards

scripts/linking.py
danielkingai2: agreed, however i can't come up with anything better...

danielkingai2: i'll just add a comment explaining

scripts/linking.py
danielkingai2: the main loop below runs over gold entities. to get the full set of predicted entities for a document, we need to loop over all predicted entities and link them. There is certainly a refactor that would allow for only calling link once for each predicted entity, but I didn't do that. And it doesn't seem to slow it down all that much.

scripts/linking.py
```diff
@@ -481,30 +481,43 @@ def get_mention_text_and_ids_by_doc(data: List[data_util.MedMentionExample],
...
-    return candidates, mention_types
+            if not use_soft_matching and span == predicted_entity:
+                candidates = filtered_batch_candidate_neighbor_ids[j]
+                mention_types = predicted_mention_types[j]
+                break
+            elif use_soft_matching:
+                overlaps = False
+                # gold span within spacy span
+                if span.start_char >= predicted_entity.start_char and span.end_char <= predicted_entity.end_char \
+                    and predicted_entity != span:
```
DeNeutoy: Doesn't this logic miss out when the predicted span is exactly equal to the gold span, due to the `predicted_entity != span` clause? I think this case should be included.

scripts/linking.py
ibeltagy: @danielkingai2 , I see. Can add a comment clarifying that part ?

scripts/linking.py
danielkingai2: No, the third if statement handles that exactly equal case, although that is quite unclear. I will make it clearer

DeNeutoy: Oh my bad

FYI I think this logic can be replaced with:

```python
    # Gold span starts inside predicted span.
if (span.start_char >= predicted_entity.start_char <= span.end_char
        # Predicted span starts inside gold span.
        or predicted_entity.start_char >= span.start_char <= predicted_entity.end_char):
	# do the overlap stuff.

```

scripts/linking.py
danielkingai2: yes, it doesn't impact anything because of how the spans are used (I think) but yes, i should set it

scripts/linking.py
danielkingai2: yeah, thats true. Iz pointed that out earlier, I'll just change it to that logic.

ibeltagy: I don't think this is working. I tried it and it gives the same result of the exact match. 
I think the overlapping logic needs to be:
```i0 = max(span_from_doc.start_char, predicted_entity.start_char)
i1 = min(span_from_doc.end_char, predicted_entity.end_char)
if i0 <= i1:
    overlap
else: 
   no overlap```

File: scripts/linking.py [hunk 6]
```diff
@@ -481,30 +481,35 @@ def get_mention_text_and_ids_by_doc(data: List[data_util.MedMentionExample],
...
-                mention_types.update(predicted_mention_types[j]) 
-
-    return candidates, mention_types
+            if not use_soft_matching and span == predicted_entity:
+                candidates.append(filtered_batch_candidate_neighbor_ids[j])
+                mention_types.append(predicted_mention_types[j])
+                mention_spans.append(predicted_entity)
+                break
+            elif use_soft_matching:
+                overlaps = False
```
ibeltagy: not used

scripts/linking.py
```diff
@@ -597,13 +615,21 @@ def eval_candidate_generation_and_linking(examples: List[data_util.MedMentionExa
                 batch_candidate_neighbor_ids = candidate_generator.generate_candidates(mention_texts, k)
 
                 filtered_batch_candidate_neighbor_ids = []
-                for candidate_neighbor_ids in batch_candidate_neighbor_ids:
+                for candidate_neighbor_ids, mention_text, predicted_entity, predicted_mention_type \
+                    in zip(batch_candidate_neighbor_ids, mention_texts, predicted_entities, predicted_mention_types):
```
ibeltagy: predicted_mention_types is not assigned with `use_gold_mentions`

scripts/linking.py
ibeltagy: I take it back. This is correct

PR #117, opened 2019-05-24
Merge candidate generator

DeNeutoy:


scispacy/candidate_generation.py
```diff
@@ -0,0 +1,307 @@
...
+        for neighbors, distances in zip(batch_neighbors, batch_distances):
+            if neighbors is None:
+                neighbors = []
+            if distances is None:
+                distances = []
+
+            concept_to_mentions: Dict[str, List[str]] = defaultdict(list)
+            concept_to_distances: Dict[str, List[float]] = defaultdict(list)
+            for neighbor_index, distance in zip(neighbors, distances):
+                mention = self.ann_concept_aliases_list[neighbor_index]
```
danielkingai2: the docstring says that `self.ann_concept_aliases_list` is `A list of strings, mapping the indices used in the ann_index to canonical UMLS ids.` although this line (and the next line) suggests it actually maps to canonical mentions. Could you clarify which is the case?

scispacy/candidate_generation.py
```diff
@@ -0,0 +1,307 @@
...
+        start_time = datetime.datetime.now()
+
+        # `ann_index.knnQueryBatch` crashes if one of the vectors is all zeros.
+        # `nmslib_knn_with_zero_vectors` is a wrapper around `ann_index.knnQueryBatch` that addresses this issue.
+        batch_neighbors, batch_distances = self.nmslib_knn_with_zero_vectors(tfidfs, k)
+        end_time = datetime.datetime.now()
+        total_time = end_time - start_time
+        if self.verbose:
+            print(f'Finding neighbors took {total_time.total_seconds()} seconds')
+        neighbors_by_concept_ids = []
```
danielkingai2: naming of this variable suggests it is a dictionary with concept ids as the key. Could you add a comment clarifying or change the name?

scispacy/candidate_generation.py
```diff
@@ -0,0 +1,307 @@
...
+    num_threads = 60  # set based on the machine
+    index_params = {'M': m_parameter, 'indexThreadQty': num_threads, 'efConstruction': construction, 'post' : 0}
+
+    print(f'No tfidf vectorizer on {tfidf_vectorizer_path} or ann index on {ann_index_path}')
+    umls_concept_aliases = list(text_to_concept_id.keys())
+
+    umls_concept_aliases = numpy.array(umls_concept_aliases)
+
+    print(f'Fitting tfidf vectorizer on {len(umls_concept_aliases)} aliases')
+    tfidf_vectorizer = TfidfVectorizer(analyzer='char_wb', ngram_range=(3, 3), min_df=10, dtype=numpy.float32)
```
danielkingai2: I know we talked about this a bunch, but I can't remember what the outcome was. Can we change this to float16?

scispacy/candidate_generation.py
```diff
@@ -0,0 +1,307 @@
...
+
+    print(f'Deleting {number_of_non_empty_tfidfs}/{total_number_of_tfidfs} aliases because their tfidf is empty')
+    # remove empty tfidf vectors, otherwise nmslib will crash
+    umls_concept_aliases = umls_concept_aliases[empty_tfidfs_boolean_flags].tolist()
+    uml_concept_alias_tfidfs = uml_concept_alias_tfidfs[empty_tfidfs_boolean_flags]
+    print(deleted_aliases)
+
+    print(f'Saving list of concept ids and tfidfs vectors to {uml_concept_aliases_path} and {tfidf_vectors_path}')
+    json.dump(umls_concept_aliases, open(uml_concept_aliases_path, "w"))
+    scipy.sparse.save_npz(tfidf_vectors_path, uml_concept_alias_tfidfs.astype(numpy.float16))
```
danielkingai2: or was it only here that we can change it to float16

DeNeutoy: I'll add a comment here, basically we can only serialize in float16

scispacy/candidate_generation.py
```diff
@@ -0,0 +1,307 @@
...
+        original_neighbours = self.ann_index.knnQueryBatch(vectors, k=k)
+
+        neighbors, distances = zip(*[(x[0].tolist(), x[1].tolist()) for x in original_neighbours])
+        neighbors = list(neighbors)
+        distances = list(distances)
+
+        # neighbors need to be convected to an np.array of objects instead of ndarray of dimensions len(vectors)xk
+        # Solution: add a row to `neighbors` with any length other than k. This way, calling np.array(neighbors)
+        # returns an np.array of objects
+        neighbors.append([])
```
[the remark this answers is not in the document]
DeNeutoy: This is actually not the case in this particular scenario, because the objects are tuples, which means that numpy unfolds the dimension, even for objects:

```
In [16]: x = numpy.array([(1,2), (3,4)], dtype=object)

In [17]: x
Out[17]:
array([[1, 2],
       [3, 4]], dtype=object)

In [18]: x = numpy.array([(1,2), (3,4), ()], dtype=object)

In [19]: x
Out[19]: array([(1, 2), (3, 4), ()], dtype=object)

In [20]:
```

I'm sure you're right that there is a good way to fix this but I couldn't find what it is...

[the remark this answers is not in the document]
danielkingai2: seems you can tell it its a float somehow, but it looks more complicated than the current hack, so 🤷‍♂

PR #125, opened 2019-06-04
Add type tree

danielkingai2:
Add the type tree to the umls class, and a get parent function to the type tree


scispacy/umls_semantic_type_tree.py
```diff
@@ -46,6 +46,20 @@ def get_children(self, node: SemanticTypeNode) -> List[SemanticTypeNode]:
             children.append(child)
             children.extend(self.get_children(child))
         return children
+    
+    def get_parent(self, node: SemanticTypeNode) -> Optional[SemanticTypeNode]:
```
DeNeutoy: Might be nice to define `parent(0) = 0` so the return type is not optional, and likely this is what you want at the end of the day?

danielkingai2: the test fixture type tree starts at level 1 :P

danielkingai2: wasn't sure if there was a "right" answer here so just went with the catch all return None if no parents

DeNeutoy: sure, if it's annoying to change the fixture no problem, lgtm

PR #127, opened 2019-06-24
Hyponym

danielkingai2:
Quick timing stats from running on MedMentions dev set:

Without hyponym pipe: 36.78 abstracts per second (23.874496698379517 seconds total)
With hyponym pipe not extended: 35.12 abstracts per second (27.79159188270569 seconds total)
With hyponym pipe extended: 31.59 abstracts per second (25.000983476638794 seconds total)


DeNeutoy:
Closing in favour of #276

PR #129, opened 2019-07-02
Make use of all Concept Types

danlou:
The 'type' field of MedMentions often includes several assignments, and you seem to have made the choice of only keeping the first type. The change I'm proposing creates additional entries ('entities') so that all type assignments are available for training, etc..
Quickly checking the difference in number of entities after this change, I'm seeing about 9% additional entities (in training set).

Just getting started with MedMentions, don't know if there may be some reason to just keep the first type, but if there isn't, this change may be of interest.

### DIFF
--- a/scispacy/data_util.py
+++ b/scispacy/data_util.py
@@ -39,9 +39,9 @@ def process_example(lines: List[str]) -> MedMentionExample:
     entities = []
     for entity_line in lines[2:]:
         _, start, end, mention, mention_type, umls_id = entity_line.split("\t")
-        mention_type = mention_type.split(",")[0]
-        entities.append(MedMentionEntity(int(start), int(end),
-                                         mention, mention_type, umls_id))
+        for mention_type_ in mention_type.split(","):
+            entities.append(MedMentionEntity(int(start), int(end),
+                                             mention, mention_type_, umls_id))
     return MedMentionExample(title, abstract, title + " " + abstract, pubmed_id, entities)
 
 def med_mentions_example_iterator(filename: str) -> Iterator[MedMentionExample]:


danielkingai2:
@DeNeutoy, I think you added this code. Is there a reason to only take the first type? I would probably suggest making `MedMentionEntity.mention_type` a list.

scispacy/data_util.py
```diff
@@ -39,9 +39,9 @@ def process_example(lines: List[str]) -> MedMentionExample:
     entities = []
     for entity_line in lines[2:]:
         _, start, end, mention, mention_type, umls_id = entity_line.split("\t")
-        mention_type = mention_type.split(",")[0]
-        entities.append(MedMentionEntity(int(start), int(end),
-                                         mention, mention_type, umls_id))
+        for mention_type_ in mention_type.split(","):
```
DeNeutoy: This is not the correct fix for this problem - we shouldn't re-append the entity with each of the types.

Instead, the `MedMentionEntity.mention_type` should have type `Union[List[str], str]`. Then you could add an option to https://github.com/allenai/scispacy/pull/129/files#diff-d747afe921fffa415e16916db28c32d8R77

which would allow this to give multiple semantic types to `MedMentionEntity`. However, I would like to preserve the current behaviour (or at least, use a more intelligent way to select a _single_ semantic type) because the NER model depends on there being a single one only and I would like the api for that to not break.

danlou: Well, then maybe you want to keep the semantic type of highest level (deeper in the hierarchy) ? At least that should be preferred for NER models that are aware of the hierarchy (as is often found in fine-grained NER).

Does scispacy support access to some sort of 'type depth' variable ?

danlou: Great, then maybe something like this would do:
`mention_type = max(mention_type.split(","), key=lambda x: umls_tree.get_node_from_id(x).level)`

I've been loading a new umls_tree though, with:
```
from scispacy.umls_semantic_type_tree import construct_umls_tree_from_tsv
umls_tree = construct_umls_tree_from_tsv("data/umls_semantic_type_tree.tsv")
```

Is `umls_tree ` initialized when loading a scispacy model, or somewhere else that we can import here?

DeNeutoy: Hi sorry for the slow reply, you can use this path from our public s3 bucket like so:
https://github.com/allenai/scispacy/blob/master/scispacy/umls_utils.py#L30

It isn't initialised when loading a model, but it's fine to just create one inside the function itself.

DeNeutoy:
Closing as stale

PR #133, opened 2019-07-03
memory efficient candidate generation 

ibeltagy:
Avoid converting from list to numpy.array and back. Now it takes less than 10GB of memory (down from 200GB)


scispacy/candidate_generation.py
```diff
@@ -290,15 +289,14 @@ def create_tfidf_ann_index(out_path: str,
 
     print(f'Finding empty (all zeros) tfidf vectors')
     empty_tfidfs_boolean_flags = numpy.array(uml_concept_alias_tfidfs.sum(axis=1) != 0).reshape(-1,)
-    deleted_aliases = umls_concept_aliases[empty_tfidfs_boolean_flags == False] # pylint: disable=singleton-comparison
```
DeNeutoy: looks like pylint doesn't like this line wrapping of the disable statement

PR #135, opened 2019-07-18
Add a couple more biological abbreviations

danielkingai2:
Came across a couple abbreviations that were annoying me by breaking sentence splitting.


scispacy/consts.py
```diff
@@ -16,4 +16,7 @@
                             "nos.",
                             "No.",
                             "Nos.",
-                            "al."]
+                            "al.",
```
DeNeutoy: (╯°□°）╯︵ ┻━┻

make it line up? 😄

danielkingai2: lol oops. will do

PR #153, opened 2019-08-14
Add `no_definition_threshold` to entity linker

### LINKED ISSUE #151: Override definition filtering for exact match
Entity linker defaults to filtering out entities that don't have definitions in UMLS, we should at least override this filtering when a mention is an exact match for a UMLS entity.

DeNeutoy:
fixes #151


tests/fixtures/umls_test_fixture.json
```diff
@@ -1 +1 @@
...lidine", "8-Aminonaphthalene", "8-Aminonaphthalene", "8 Aminonaphthalene", "1-Aminonaphthalene", "1-Aminonaphthalene", "1 Aminonaphthalene", "1-Naphthalenamine", "1-Naphthalenamine", "1-Naththylamine", "1-Naththylamine", "a- Naphthylamine", "a- Naphthylamine", "a-Naphthylamine", "1-Naththylamine (substance)", "1-Naphthylamine [Chemical/Ingredient]"], "types": ["T109", "T131"], "canonical_name": "1-Naphthylamine", "definition": "A suspected industrial carcinogen (and listed as such by OSHA). Its N-hydroxy metabolite is strongly carcinogenic and mutagenic."}, {"concept_id": "C0000103", "aliases": ["1-Naphthylisothiocyanate", "1 Naphthylisothiocyanate", "alpha-Naphthylisothiocyanate", "alpha-Naphthylisothiocyanate", "alpha Naphthylisothiocyanate", "Naphthalene, 1-isothiocyanato-", "Naphthalene, 1-isothiocyanato-", "1-Naphthylisothiocyanate [Chemical/Ingredient]"], "types": ["T109", "T130", "T131"], "canonical_name": "1-Naphthylisothiocyanate", "definition": "A tool for the study of liver damage which causes bile stasis and hyperbilirubinemia acutely and bile duct hyperplasia and biliary cirrhosis chronically, with changes in hepatocyte function. It may cause skin and kidney damage."}]
```
DeNeutoy: Apologies for this diff, I just hacked in an entity that was similar-ish to another one so that I could test for the definition threshold thing.

scispacy/umls_linking.py
```diff
@@ -96,8 +96,8 @@ def __call__(self, doc: Doc) -> Doc:
             for cand in candidates:
                 score = max(cand.similarities)
                 if (self.filter_for_definitions and
-                    self.umls.cui_to_entity[cand.concept_id].definition is None and
-                    score < self.no_definition_threshold):
+                            self.umls.cui_to_entity[cand.concept_id].definition is None and
```
danielkingai2: lol

tests/test_candidate_generation.py
```diff
@@ -31,7 +31,7 @@ def test_candidate_generation(self):
         results = candidate_generator(['(131)I-Macroaggregated Albumin'], 10)
 
         canonical_ids = [x.concept_id for x in results[0]]
-        assert canonical_ids == ['C0000005', 'C0000102', 'C0000084']
```
danielkingai2: wait why did this bit change? You added 'C0000015' right? Why did other stuff change?

DeNeutoy: Ummm, i'm going to say this is a quirk of the ANN thing?

danielkingai2: hmm, perhaps. thats a bit odd given the size of the kb. maybe it performs strangely on small kbs

danielkingai2: hmm, perhaps. thats a bit odd given the size of the kb. maybe it performs strangely on small kbs

DeNeutoy: I think this is the case - the index in the tests is particularly unstable because the TF-IDF vectorizer contains very few char-3grams, meaning additions of any new aliases or entities could substantially change similarity scores between entities. (I checked this by adding some random aliases to the entity I added and the results changed again.) This isn't ideal for a test but I don't have a good alternative either....

DeNeutoy: So yeah, I think it's not the ANN thing, but more that similarity between TF-IDF vectors induced from a tiny corpora doesn't make too much sense.

danielkingai2: seems fine 👍

PR #164, opened 2019-09-17
Update README.md

danielkingai2:
Conda download link was broken


danielkingai2:
Alright, teamcity is fully off now

PR #168, opened 2019-10-03
Pin scispacy to 2.1.x

danielkingai2:
The current models are not compatible with spacy 2.2. Pinning the scispacy version until we can upgrade to 2.2 and release new models.

### DIFF
--- a/scispacy/abbreviation.py
+++ b/scispacy/abbreviation.py
@@ -191,4 +191,4 @@ class AbbreviationDetector:
             # Clean up the global matcher.
             self.global_matcher.remove(key)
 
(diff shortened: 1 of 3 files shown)


scispacy/abbreviation.py
```diff
@@ -191,4 +191,4 @@ def find_matches_for(self,
             # Clean up the global matcher.
             self.global_matcher.remove(key)
 
-        return [(k, v) for k, v in all_occurences.items()]
```
DeNeutoy: RIP

danielkingai2:
Oops, in my head I merged this. Thanks

PR #169, opened 2019-10-15
Fixes required for upgrading to spacy 2.2.1

danielkingai2:
- add spacy-lookups-data package
- update requirements and setup.py with 2.2.1
- remove overlapping entities from the medmentions ner data
- superficial changes to parser/tagger training script
- fix umls linking test because of spacy entities api changing

Note: retraining the models is separate from this PR, and currently requires this branch of my fork of spacy to fix a bug in the current release: https://github.com/danielkingai2/spaCy/tree/most_similar_bug


scispacy/data_util.py
```diff
@@ -61,18 +61,69 @@ def med_mentions_example_iterator(filename: str) -> Iterator[MedMentionExample]:
...
-        spacy_format_entities = [(x.start, x.end, x.mention_type) for x in example.entities]
-        examples.append((example.text, {"entities": spacy_format_entities}))
-
-    return examples
+    entity_index = 0
+    spacy_format_entities_without_overlap = []
+    stack = []
+    stack_start = 0
+    stack_end = 0
+    while True:
```
DeNeutoy: This while loop should be a for loop, because you don't modify the `entity_index` variable apart from at the end of the loop. You can then simplify the logic in the loop by handling the `current_entity=None` case outside the body of the loop.

scispacy/data_util.py
```diff
@@ -61,18 +61,69 @@ def med_mentions_example_iterator(filename: str) -> Iterator[MedMentionExample]:
...
+        else:
+            # if we are after the end of the list, just go through the last iteration to dump the stack
+            current_entity = None
 
+        if current_entity is not None and len(stack) == 0:
+            stack.append(current_entity)
+            stack_start = current_entity_start
+            stack_end = current_entity_end
+        else:
+            if current_entity is not None \
```
DeNeutoy: Hmm this logic is pretty confusing.

can't we do it using a scan over the entities?

```python
sorted_by_start_then_end = sorted(entities, key= lambda x: (x[0], x[1]))
non_overlapping = [sorted_by_start_then_end[0]]

for entity in sorted_by_start_then_end[1:]:
    start, end, _ = entity
	prev_start, prev_end, _ = non_overlapping[-1]

	# case 1: strict non-overlapping.
	if start >= prev_end:
		non_overlapping.append(entity)
	
	# case 2: spans are the same, or new entity is 
	# contained within previous. So skip
	elif start >= prev_start and end <= prev_end:
		continue
	
	# case 3: overlapping, so we need to check which is longer.
	elif start > prev_start and end > prev_end:
		
		# this entity is bigger - pop previous, add this one instead.
		if end - start > prev_end - prev_start:
			non_overlapping.pop()
			non_overlapping.append(entity)
```

Maybe i'm missing some assumption which will make this not work?

scispacy/data_util.py
```diff
@@ -61,18 +61,69 @@ def med_mentions_example_iterator(filename: str) -> Iterator[MedMentionExample]:
...
+                    and len(range(max(current_entity_start, stack_start), min(current_entity_end, stack_end))) > 0:
```
DeNeutoy: this len(range(...)) can just be a numeric comparison

PR #171, opened 2019-10-17
Fix bad bug in merging pysbd segments

danielkingai2:
Thanks for finding @isabelcachola, hopefully there aren't any more :)

Edit: This pr has become a couple of hacks to fall back on the default sentence splitter if pysbd does something weird or fails. This is the first stage of testing out this pipe as the sentence splitter.

Edit Edit: known pysbd issues have been fixed, so i've removed the fall backs


scispacy/custom_sentence_segmenter.py
```diff
@@ -35,6 +40,7 @@ def combined_rule_sentence_segmenter(doc: Doc) -> Doc:
     current_segment = segments[segment_index]
     built_up_sentence = ""
     for i, token in enumerate(doc):
+        # print(current_segment, "BREAK", built_up_sentence, "BREAK", token)
```
DeNeutoy: remove commented code

scispacy/custom_sentence_segmenter.py
```diff
@@ -26,7 +31,16 @@ def combined_rule_sentence_segmenter(doc: Doc) -> Doc:
 
     @param doc: the spaCy document to be annotated with sentence boundaries
     """
-    segmenter = pysbd.Segmenter(language="en", clean=False)
+    # pysbd doesn't always like short strings
```
nipunsadvilkar: @danielkingai2 : Can you share few examples of such cases?

danielkingai2: @nipunsadvilkar the one i noticed was that it turned `.` into ``. And it seemed harmless to not split segments on one single token documents so I just special cased it.

nipunsadvilkar: I didnt understand it fully. Can you provide an example, please?

danielkingai2:
```
>>> import pysbd
>>> segmenter = pysbd.Segmenter(language='en', clean=False)
>>> segmenter.segment('.')
[]
>>> 
```

nipunsadvilkar: Got it. Thanks 👍

nipunsadvilkar: bdw I have fixed https://github.com/nipunsadvilkar/pySBD/issues/27 bug.

danielkingai2:
Current status is that all tests pass without the fall backs on the normal sentence splitter. Leaving them in for now (without merging), hoping that @isabelcachola can rerun her job and see if there are any more edge cases that output warnings, or if we can remove the fall backs.

danielkingai2:
@DeNeutoy all the hacky fallbacks have been removed because all the known pysbd bugs were fixed. I'll merge it in this (better) state once CI passes unless you object to something

PR #187, opened 2019-12-11
add links to the demo

DeNeutoy:


tests/test_umls_linking.py
```diff
@@ -23,6 +23,7 @@ def setUp(self):
     def test_naive_entity_linking(self):
         text = "There was a lot of Dipalmitoylphosphatidylcholine."
         doc = self.nlp(text)
+        doc.ents = (doc[5:6],)
```
danielkingai2: oh did ner change?

DeNeutoy: Yeah, the new spacy 2.2.5 models don't naturally detect that long name as an entity anymore, i'm not really sure why I tested it using the default models in the first place...

PR #188, opened 2019-12-27
Avoid spanning over contiguous sentences in abbreviation resolver

dnanhkhoa:
In this PR, I fixed an issue in AbbreviationDetector that a long-form span can span over contiguous sentences.

You can check this snippet

```py
import spacy

from scispacy.abbreviation import AbbreviationDetector

nlp = spacy.load("en_core_sci_sm")

# Add the abbreviation pipe to the spacy pipeline.
abbreviation_pipe = AbbreviationDetector(nlp)
nlp.add_pipe(abbreviation_pipe)

doc = nlp(
"""
Logical design of an anti - cancer agent targeting the plant homeodomain in Pygopus2 .
Pygopus2 ( Pygo2 ) is a component of the Wnt signaling pathway , which is required for β - catenin mediated transcription .
"""
)

print("Abbreviation", "\t", "Definition")
for abrv in doc._.abbreviations:
    print(f"`{abrv}` \t ({abrv.start}, {abrv.end}) `{abrv._.long_form}`")
```

Before the fix, the result is:
```
`Pygo2`          (19, 20) `agent targeting the plant homeodomain in Pygopus2 .
Pygopus2`
```

After:
```
Abbreviation     Definition
`Pygo2`          (19, 20) `Pygopus2`
```


dnanhkhoa:
After checking, I failed the test case due to the broken sentence splitter. It splits the sentence "1-Methyl-4-phenylpyridinium (MPP+) is an abbreviation which doesn't exist in the baby index." into 2 pieces: "1-Methyl-4-phenylpyridinium" and "(MPP+) is an abbreviation which doesn't exist in the baby index.". This is weird and I don't know what I should do for this case. Please let me know your idea. Thanks

DeNeutoy:
@dnanhkhoa thanks for the PR! It looks like there is some bug in your implementation as this test (`TestAbbreviationDetector.test_find`) is failing: https://github.com/allenai/scispacy/runs/364816000#step:3:625 . 

I don't think this is due to the sentence splitting, but I could be wrong. Maybe if you fix this one, the other one will also resolve?

PR #194, opened 2020-02-10
Add ontology code-linking to UmlsEntity (e.g., ICD10CM, Loinc, RxNorm, etc)

snosrap:
This adds a dictionary of `codes` to each UmlsEntity, allowing expressions like: `linker.umls.cui_to_entity[cui].codes['ICD10CM']` to get a list of ICD10CM codes. The code key/values are defined by the SAB/CODE columns of MRCONSO.

Example:

```
    {
        "concept_id": "C0000005",
        "aliases": [
            "(131)I-MAA"
        ],
        "types": [
            "T116"
        ],
        "codes": { // <-- new
            "MSH": [
                "D012711",
                "D012711"
            ]
        },
        "canonical_name": "(131)I-Macroaggregated Albumin"
    },
```

This admittedly makes the JSON file a lot bigger.

Each dictionary value returns a **list** of `codes` for the key'd ontology (which likely includes duplicates). I took this approach because it seemed similar to your `aliases` field and allows for more processing downstream. A Counter dictionary would be another approach.

Open to suggestions.

### DIFF
--- a/scispacy/umls_utils.py
+++ b/scispacy/umls_utils.py
@@ -15,6 +15,7 @@ class UmlsEntity(NamedTuple):
     canonical_name: str
     aliases: List[str]
     types: List[str]
+    codes: Dict[str, List[str]]
     definition: Optional[str] = None
 
     def __repr__(self):
@@ -148,6 +149,7 @@ def read_umls_concepts(meta_path: str, concept_details: Dict):
                     "concept_id": concept_id,
                     "aliases": [],
                     "types": [],
+                    "codes": defaultdict(list)
                 }
 
             concept_name = concept["STR"]
@@ -168,6 +170,7 @@ def read_umls_concepts(meta_path: str, concept_details: Dict):
                     "canonical_name"
                 ] = concept_name  # set as canonical name
 
+            concept_details[concept_id]['codes'][concept["SAB"]].append(concept["CODE"])
 
 def read_umls_types(meta_path: str, concept_details: Dict):
     """


DeNeutoy:
Hi, thanks for the PR!

How much bigger does this make the json file? I'm not sure we'll accept this change, because you could just create this mapping separately and use it (i.e it is not actually required to train the linker). Hopefully that makes sense?

snosrap:
I can't compare exactly (since I only had access to MRCONSO, not the other files), but maybe 800MB or so? So way bigger.

Totally reasonable to map things separately for exactly the reason you suggest.

That said, I think there could be some heuristic value to having access to code/ontology data in the linker -- the count of mapped codes could be a signal for CUI importance/relevance, or you could link in a way that gives preference to a particular codeset/ontology (e.g., require ICD10CM linkages for entities in the "diagnosis vector space").

I'm trying to come up with a good example use case for this heuristic, but all the entity-CUI linking misses I'm seeing are mostly from common clinical abbreviations (e.g., "CAD" for coronary artery disease, "DM2" for type 2 diabetes) that I don't think would be solved by using an ontology-aware linker (since those abbreviations aren't in MRCONSO in a useful way). I'll keep looking for a good example of where this would be useful, but for now it's just a hunch.

DeNeutoy:
Gotcha - FYI we have an abbreviation detector, which if you run before the linker, will help abbreviation linking performance - you might want to try it out! 

I'm going to close this, as I think until we have a concrete use-case, it might be best to leave it out.

PR #209, opened 2020-03-27
Tweak short form filter of abbreviation detector and apply it in both cases

### LINKED ISSUE #161: Special cases in abbreviation algorithm
There seem to be some low hanging fruit for making the abbreviation detector work better on paper body text. Storing examples here in case we want to make some easy fixes at some point:

```
(text --- short form --- long form)
("H2)]+(14)s.t. (1), (4).Similarly" --- 1 --- H2)]+(14)s.t.)
(".(21)In (21), λ" --- 21 --- .(21)In)
("map expX (·) : R" --- . --- map expX)
("0,(3)with the following data: (3-i) (q̄" --- 3-i --- 0,(3)with the following data:)
( "Φg(h),ThΦg(v) ) , (h, v)" --- h, v --- Φg(h),ThΦg(v) ) ,)
("dimension;(S-iii) The optimal control problem obtained in (S-ii) is con-verted" --- S-ii --- dimension;(S-iii) The optimal control problem obtained in)
("z), πut (z)) )" --- z) --- z), πut)
("is equivalent to (iv) of Theorem" --- iv --- is equivalent to)
("or to fork.Users work more on their repositories (owners) than on" --- owners --- or to fork.Users work more on their repositories)
("repositories he/she already worked with or from previous collaborators. Nevertheless, 88% of the first action of users to a repository (repository discovery) is" --- repository discovery --- repositories he/she already worked with or from previous collaborators. Nevertheless, 88% of the first action of users to a repository)
```

and so on and so forth

danielkingai2:
Fixes #161 a couple cases are xfailed because i don't see an obvious way to exclude them (i.e. they aren't the result of an obvious bug, but a failure of the algorithm)

### DIFF
--- a/scispacy/abbreviation.py
+++ b/scispacy/abbreviation.py
@@ -120,6 +120,7 @@ def short_form_filter(span: Span) -> bool:
     # All words are between length 2 and 10
     if not all([2 <= len(x) < 10 for x in span]):
         return False
+
     # At least 50% of the short form should be alpha
     if (sum([c.isalpha() for c in span.text]) / len(span.text)) < 0.5:
         return False
--- a/scispacy/abbreviation.py
+++ b/scispacy/abbreviation.py
@@ -96,26 +96,36 @@ def filter_matches(
             # Long form is inside the parens.
             # Take one word before.
             short_form_candidate = doc[start - 2 : start - 1]
-            if short_form_filter(short_form_candidate):
-                candidates.append((doc[start:end], short_form_candidate))
+            long_form_candidate = doc[start:end]
         else:
             # Normal case.
             # Short form is inside the parens.
+            short_form_candidate = doc[start:end]
+
             # Sum character lengths of contents of parens.
-            abbreviation_length = sum([len(x) for x in doc[start:end]])
+            abbreviation_length = sum([len(x) for x in short_form_candidate])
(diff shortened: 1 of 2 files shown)


scispacy/abbreviation.py
```diff
@@ -96,26 +96,36 @@ def filter_matches(
...
+        if short_form_filter(short_form_candidate):
+            candidates.append((long_form_candidate, short_form_candidate))
+
     return candidates
 
 
 def short_form_filter(span: Span) -> bool:
     # All words are between length 2 and 10
-    if not all([2 < len(x) < 10 for x in span]):
+    if not all([2 <= len(x) < 10 for x in span]):
```
DeNeutoy: I believe strictly greater than 2 here is motivated by one of your xfail cases, namely that references, citations etc are often single or 2 char identifiers, like `(1), [2], (iv)` etc. But fine if it catches some case you need!

danielkingai2: yeah, the thing is that previously the short form filter was only being applied in the <short> (<long>) case. When applied in both cases, it even fails one of the tests we have, which contains a two char abbreviation. In general, I think there can definitely be short forms that are two chars. I think whatever extra ones show up like (iv) will be easier to filter in an end use case than it would be to capture missing two char abbreviations so i'm going to go ahead and merge this.

PR #223, opened 2020-05-05
Fix space issue for long form

danielkingai2:


scispacy/abbreviation.py
```diff
@@ -61,17 +61,17 @@ def find_abbreviation(
         long_index -= 1
         short_index -= 1
 
-    # If we complete the string, we end up with -1 here,
-    # but really we want all of the text.
-    long_index = max(long_index, 0)
+    # The last subtraction takes us on to a space character
```
DeNeutoy, line 64: This change works because we only ever reach this code if we do complete the string - it's also not quite the case that it takes us onto a space, because here we are operating over words which don't have spaces. I think the original comment is a little clearer?

scispacy/abbreviation.py
```diff
@@ -61,17 +61,17 @@ def find_abbreviation(
...
+    long_index += 1
 
     # Now we know the character index of the start of the character span,
     # here we just translate that to the first token beginning after that
     # value, so we can return a spaCy span instead.
     word_lengths = 0
     starting_index = None
     for i, word in enumerate(long_form_candidate):
-        word_lengths += len(word)
+        # need to add 1 for the space characters
```
DeNeutoy, line 73: Hmm, I'm not sure about this. Is it guaranteed that spacy represents token boundaries with a single whitespace, rather than multiple? Because here you assume it's always one. I think this was working before because we rarely encounter a long form candidate which contains single letter words, and then it would still work if we are off by one. 

I think the right thing to do here might be to use the `text_with_ws` attribute from spacy when we construct the long form, so that we match their whitespace exactly.

danielkingai2: I think you're right that `text_with_ws` is the right thing to use, but it was more than an off by one thing before, because the `long_form_candidate` could contain multiple spaces. Before there was some lucky math combining the off by one from above and the number of words we take before the abbreviation. For example:

```
In [24]: text = "this text has a bunch more words this text has five words before the Newsss Abbreviation Text Under All (NATURAL)"

In [25]: nlp(text)._.abbreviations[0]._.long_form
Out[25]: Abbreviation Text Under All
```

scispacy/abbreviation.py
danielkingai2: we're not operating over text without whitespace here right? because the long/short form are defined above as space joined text tokens.

danielkingai2: which makes me notice that the indexing is actually over these single space joined tokens, and that will need to be unified with the below code, rather than just using `text_with_ws`

danielkingai2: I think to match this, below I will add `len(word.text) + 1` for each word

danielkingai2: oh actually when there are multiple spaces, spacy give the space its own token.

DeNeutoy: Right, but if we "go off the front" of the string, it won't be a whitespace char (it will refer to nothing as we are just indexing a string and not the doc), which made your comment a little confusing for me. It's a small thing.

danielkingai2: Oh i see, let me clarify that there are two cases there.

DeNeutoy: Yeah, I think this whole thing would be easier if we moved it to operate on tokens rather than strings, it would reduce how annoying this is in the edge cases, you're right

scispacy/abbreviation.py
danielkingai2: Ok, updated to use `text_with_ws` and added a test that has multiple white spaces in the string.

danielkingai2:
@DeNeutoy CI is running now, see if you think this is correct? (even if maybe it should be rewritten to just operate on tokens)

PR #233, opened 2020-05-18
Linking refactor

DeNeutoy:
This is an attempt to make it easier to load other KBs trained via the same mechanism as a pipeline. 

Changes:

- `UmlsEntity` -> `Entity`
- `types` argument to `UmlsEntity` is now optional, as not every KB will have types
- Abstract `UmlsKnowledgeBase` into `KnowledgeBase`, which doesn't hold the semantic type tree of UMLS.
- Wrap up all the linker paths into a `LinkerPaths` namedtuple, so we can reference groups of them by name
- Pass a `name` arg to the linker and candidate generator, which is enough to construct the various pre-defined linkers we have.

~~I'll actually add the MESH linker in a different PR to keep this one manageable.~~
edit: Yolo it's not much extra code, i'll just add it here.

Once you've reviewed i'll add the data and stuff to s3!


scispacy/candidate_generation.py
```diff
@@ -11,13 +11,39 @@
...
+from scispacy.umls_utils import KnowledgeBase, UmlsKnowledgeBase, MeshKnowledgeBase
 
-DEFAULT_PATHS = {
-    "ann_index": "https://s3-us-west-2.amazonaws.com/ai2-s2-scispacy/data/linking_model/nmslib_index.bin",
-    "tfidf_vectorizer": "https://s3-us-west-2.amazonaws.com/ai2-s2-scispacy/data/linking_model/tfidf_vectorizer.joblib",  # noqa
-    "tfidf_umls_vectors": "https://s3-us-west-2.amazonaws.com/ai2-s2-scispacy/data/linking_model/tfidf_vectors_sparse.npz",  # noqa
-    "concept_aliases_list": "https://s3-us-west-2.amazonaws.com/ai2-s2-scispacy/data/linking_model/concept_aliases.json",  # noqa
+
+class LinkerPaths(NamedTuple):
+    ann_index: str
```
danielkingai2, line 18: Could you add comments here stating what kind of file each of these things are and what should be in them?

scispacy/candidate_generation.py
```diff
@@ -11,13 +11,39 @@
...
+
+
+UmlsLinkerPaths = LinkerPaths(
+    ann_index="https://s3-us-west-2.amazonaws.com/ai2-s2-scispacy/data/linking_model/nmslib_index.bin",
+    tfidf_vectorizer="https://s3-us-west-2.amazonaws.com/ai2-s2-scispacy/data/linking_model/tfidf_vectorizer.joblib",  # noqa
+    tfidf_vectors="https://s3-us-west-2.amazonaws.com/ai2-s2-scispacy/data/linking_model/tfidf_vectors_sparse.npz",  # noqa
+    concept_aliases_list="https://s3-us-west-2.amazonaws.com/ai2-s2-scispacy/data/linking_model/concept_aliases.json",  # noqa
+)
+
+MeshLinkerPaths = LinkerPaths(
```
danielkingai2, line 31: 99% sure you did this intentionally and will upload to s3 later, but commenting anyway

scispacy/candidate_generation.py
```diff
@@ -109,39 +131,54 @@ class CandidateGenerator:
...
         tfidf_vectorizer: TfidfVectorizer = None,
         ann_concept_aliases_list: List[str] = None,
-        umls: UmlsKnowledgeBase = None,
+        umls: KnowledgeBase = None,
         verbose: bool = False,
         ef_search: int = 200,
+        name: str = None,
     ) -> None:
 
+        if name is not None and any(
```
danielkingai2, line 157: This parameter setup seems a bit confusing, but I don't have a suggestion for easy way to fix it so i guess its fine.

scispacy/candidate_generation.py
```diff
@@ -109,39 +131,54 @@ class CandidateGenerator:
     ann_concept_aliases_list: List[str]
         A list of strings, mapping the indices used in the ann_index to possible UMLS mentions.
         This is essentially used a lookup between the ann index and actual mention strings.
-    umls: UmlsKnowledgeBase
```
danielkingai2, line 112: should this still be called umls? or something more generic?

scispacy/candidate_generation.py
```diff
@@ -109,39 +131,54 @@ class CandidateGenerator:
...
-            cached_path(DEFAULT_PATHS["tfidf_vectorizer"])
+            cached_path(linker_paths.tfidf_vectorizer)
         )
         self.ann_concept_aliases_list = ann_concept_aliases_list or json.load(
-            open(cached_path(DEFAULT_PATHS["concept_aliases_list"]))
+            open(cached_path(linker_paths.concept_aliases_list))
         )
 
-        self.umls = umls or UmlsKnowledgeBase()
+        self.umls = umls or DEFAULT_KNOWLEDGE_BASES[name]()
```
danielkingai2, line 181: should this still be called umls or something more generic?

scispacy/candidate_generation.py
```diff
@@ -267,19 +304,16 @@ def __call__(
 
 
 def create_tfidf_ann_index(
-    out_path: str, umls: UmlsKnowledgeBase = None
+    out_path: str, umls: KnowledgeBase = None
```
danielkingai2, line 307: should this still be called umls?

scispacy/candidate_generation.py
```diff
@@ -267,19 +304,16 @@ def __call__(
...
 
-    Warning: Running this function on the whole of UMLS requires ~ 200GB of RAM ...
-    TODO: Make this not take 200GB of RAM.
-
     Parameters
     ----------
     out_path: str, required.
         The path where the various model pieces will be saved.
-    umls : UmlsKnowledgeBase, optional.
+    umls : KnowledgeBase, optional.
```
danielkingai2, line 316: should this still be called umls?

scispacy/umls_utils.py
```diff
@@ -36,49 +36,75 @@ def __repr__(self):
...
+    ):
+
+        super().__init__(file_path)
+
         self.semantic_type_tree: UmlsSemanticTypeTree = construct_umls_tree_from_tsv(
             types_file_path
         )
 
 
+class MeshKnowledgeBase(KnowledgeBase):
```
danielkingai2, line 101: I believe mesh terms have a tree as well, if one of us is able to find that tree somewhere on the internet it would be nice to package it here.

scispacy/linking.py
```diff
@@ -0,0 +1,116 @@
...
+    entities above a given threshold.
+
+
+    This class sets the `._.umls_ents` attribute on spacy Spans, which consists of a
+    List[Tuple[str, float]] corresponding to the UMLS concept_id and the associated score
+    for a list of `max_entities_per_mention` number of entities.
+
+    You can look up more information for a given id using the umls attribute of this class:
+
+    print(linker.umls.cui_to_entity[concept_id])
```
danielkingai2, line 25: because the attribute is still called umls, this is a bit confusing now that mesh is an option. I'm guessing you left it as umls throughout for backward compat?

scispacy/linking.py
```diff
@@ -0,0 +1,116 @@
...
+        resolve_abbreviations: bool = True,
+        k: int = 30,
+        threshold: float = 0.7,
+        no_definition_threshold: float = 0.95,
+        filter_for_definitions: bool = True,
+        max_entities_per_mention: int = 5,
+        name: str = None,
+    ):
+
+        Span.set_extension("umls_ents", default=[], force=True)
```
danielkingai2, line 73: same question about keeping the "umls" naming

scispacy/linking.py
```diff
@@ -0,0 +1,116 @@
...
+
+        self.candidate_generator = candidate_generator or CandidateGenerator(name=name)
+        self.resolve_abbreviations = resolve_abbreviations
+        self.k = k
+        self.threshold = threshold
+        self.no_definition_threshold = no_definition_threshold
+        self.umls = self.candidate_generator.umls
```
danielkingai2, line 80: umls again

scispacy/candidate_generation.py
DeNeutoy: Yeah, the idea is you _either_ provide a name (which we know about), or you provide all the components yourself.

scispacy/candidate_generation.py
DeNeutoy: yeahhh this is kind of annoying. I could change it to `kb` everywhere, leaving `umls` there also for a couple of releases or something?

scispacy/umls_utils.py
DeNeutoy: They do, but the only reason we have the tree in there in the first place was because we were messing around with changing the granularity of the types for the NER model, which turned out to not be super useful. Also, the MESH KB lives in this hyper complex sparql database and I really don't want to figure out how to extract the tree haha

scispacy/candidate_generation.py
danielkingai2: Yeah, the error message is probably good enough, but maybe add a note in the docstring about this point.

scispacy/candidate_generation.py
danielkingai2: You have more experience with changes like this from allennlp, so happy to defer to you, but that sounds reasonable to me. add a deprecation warning when using umls or something maybe

scispacy/umls_utils.py
danielkingai2: lol fair enough. i do think that one of the pieces of value we add here is easy access to these kbs that are pretty complicated to figure out how to access (and people do use the type tree, or at least the levels of the types). so if its not too much work, it'd be great to extract and add, if its difficult, im fine letting it go.

DeNeutoy: That's true - I'll add a TODO, maybe we can do it later. Mesh has a much better browser for viewing the entities, e.g https://meshb.nlm.nih.gov/record/ui?ui=D002430

scispacy/candidate_generation.py
DeNeutoy: cool, will do

scispacy/umls_utils.py
danielkingai2: maybe just open an issue, and if someone has a compelling use case we can add it, or maybe they will add it :)

DeNeutoy:
Ok, I moved from `umls` to `kb`, including for the spacy span annotations, but I left umls_ents there too for the time being. We can remove it in some later version.

PR #246, opened 2020-07-07
New linkers

DeNeutoy:
- use jsonl files for new linkers, as they are easier to work with
- add paths to new linkers: `rxnorm`, `go`, `hpo` and updated `umls` linker.


README.md
```diff
@@ -127,10 +127,17 @@ for abrv in doc._.abbreviations:
 ```
 ### EntityLinker
 
-The `EntityLinker` is a SpaCy component which performs linking to a knowledge base. Currently, the
-Unified Medical Language System and the Medical Subject Headings (MESH) are supported.
-The linker simply performs a string overlap search on named entities,
-comparing them with a knowledge base of 2.7 million concepts using an approximate nearest neighbours search.
+The `EntityLinker` is a SpaCy component which performs linking to a knowledge base. The linker simply performs
+a string overlap based search on named entities, comparing them with a knowledge base of 2.7 million concepts
```
danielkingai2, line 131: its a little unclear what the 2.7 number is referring to here because there are 5 options below. probably mention the 2.7 refers to the default, umls option.

danielkingai2, line 131: its a little unclear what the 2.7 number is referring to here because there are 5 options below. probably mention the 2.7 refers to the default, umls option.

scispacy/umls_utils.py
```diff
@@ -37,7 +37,7 @@ def read_umls_file_headers(meta_path: str, filename: str) -> List[str]:
     return None
 
 
-def read_umls_concepts(meta_path: str, concept_details: Dict):
+def read_umls_concepts(meta_path: str, concept_details: Dict, source: str = None):
```
danielkingai2, line 40: add source arg to the docstring?

PR #276, opened 2020-10-12
Spacy Matcher for hearst patterns

DeNeutoy:
A spacy matcher for hearst patterns, built off of #127 / Steve Anning's work. We need to get approval from him that we can use his converted patterns before we merge this, but it is functional.

### DIFF
--- a/scispacy/hearst_patterns.py
+++ b/scispacy/hearst_patterns.py
@@ -1,0 +1,527 @@
+hypernym = {"POS": {"IN": ["NOUN", "PROPN", "PRON"]}}
+hyponym = {"POS": {"IN": ["NOUN", "PROPN", "PRON"]}}
+punct = {"IS_PUNCT": True, "OP": "?"}
+det = {"ORTH": "*", "OP": "*"}
+
+BASE_PATTERNS = [
+    # '(NP_\\w+ (, )?such as (NP_\\w+ ?(, )?(and |or )?)+)', 'first'
+    {
+        "label": "such_as",
+        "pattern": [hypernym, punct, {"LEMMA": "such"}, {"LEMMA": "as"}, det, hyponym],
+        "position": "first",
+    },
+    # '(NP_\\w+ (, )?include (NP_\\w+ ?(, )?(and |or )?)+)', 'first'
+    {
+        "label": "include",
+        "pattern": [hypernym, punct, {"LEMMA": "include"}, det, hyponym],
+        "position": "first",
+    },
+    # '(NP_\\w+ (, )?especially (NP_\\w+ ?(, )?(and |or )?)+)', 'first'
+    {
+        "label": "especially",
+        "pattern": [hypernym, punct, {"LEMMA": "especially"}, det, hyponym],
+        "position": "first",
+    },
+    # '((NP_\\w+ ?(, )?)+(and |or )?other NP_\\w+)', 'last'
+    {
+        "label": "other",
+        "pattern": [
+            det,
+            hyponym,
+            punct,
+            {"LEMMA": {"IN": ["and", "or"]}},
+            {"LEMMA": "other"},
+            hypernym,
+        ],
+        "position": "last",
+    },
+]
+
+EXTENDED_PATTERNS = [
+    # '(NP_\\w+ (, )?which may include (NP_\\w+ ?(, )?(and |or )?)+)', 'first'
+    {
+        "label": "which_may_include",
+        "pattern": [
+            hypernym,
+            punct,
+            {"LEMMA": "which"},
+            {"LEMMA": "may"},
+            {"LEMMA": "include"},
+            det,
+            hyponym,
+        ],
+        "position": "first",
+    },
+    # '(NP_\\w+ (, )?which be similar to (NP_\\w+ ? (, )?(and |or )?)+)', 'first'
+    {
+        "label": "which_be_similar_to",
+        "pattern": [
+            hypernym,
+            punct,
+            {"LEMMA": "which"},
+            {"LEMMA": "be"},
+            {"LEMMA": "similar"},
+            {"LEMMA": "to"},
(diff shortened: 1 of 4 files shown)


tests/test_hyponym_detector.py
```diff
@@ -0,0 +1,49 @@
...
+
+        doc = self.nlp("SARS, or other coronaviruses, are bad.")
+        assert doc._.hearst_patterns == [("other", doc[4:5], doc[0:1])]
+        doc = self.nlp("Coronaviruses, including SARS and MERS, are bad.")
+        assert doc._.hearst_patterns == [
+            ("include", doc[0:1], doc[3:4]),
+            ("include", doc[0:1], doc[5:6])
+        ]
+
+    def test_find_noun_compound_head(self):
```
danielkingai2, line 33: is this stuff necessary vs just using noun chunks?

scispacy/hyponym_detector.py
```diff
@@ -0,0 +1,155 @@
...
+                hyponym = doc[start]
+            else:
+                # An inelegent way to deal with the "such_NOUN_as pattern"
+                # since the first token is not the hypernym.
+                if doc[start].lemma_ == "such":
+                    start += 1
+                hypernym = doc[start]
+                hyponym = doc[end - 1]
+
+            hypernym = self.find_noun_compound_head(hypernym)
```
danielkingai2, line 122: are we just appending the head here? shouldn't this be the whole noun chunk?

scispacy/hyponym_detector.py
```diff
@@ -0,0 +1,155 @@
...
+            # hypernym recorded as True and list of hyponyms created
+            hypernym._.is_hypernym = True
+            hypernym._.hyponyms.append(hyponym)
+
+            # hyponym recorded as True and its hypernym is recorded
+            hyponym._.is_hyponym = True
+            hyponym._.hypernym = hypernym
+
+            # iterate over conjunct list attached to hyponym
+            for token in hyponym.conjuncts:
```
danielkingai2, line 134: no experience using this property, but I'll assume this is reasonable

tests/test_hyponym_detector.py
danielkingai2: ^probably should have put this comment in the code, but i read the test first

scispacy/hearst_patterns.py
```diff
@@ -0,0 +1,526 @@
+hypernym = {"POS": {"IN": ["NOUN", "PROPN", "PRON"]}}
```
danielkingai2, line 1: I see, I now remember that the reason we didn't implement this with a matcher before was because there was no way to match noun chunks without using the dep matcher. I suppose this is fine though

scispacy/hearst_patterns.py
```diff
@@ -0,0 +1,526 @@
...
+hyponym = {"POS": {"IN": ["NOUN", "PROPN", "PRON"]}}
+punct = {"IS_PUNCT": True, "OP": "?"}
+det = {"ORTH": "*", "OP": "*"}
+
+BASE_PATTERNS = [
```
danielkingai2, line 6: didn't check all the patterns, assume they're fine

danielkingai2 (approved):
A couple questions, looks fine overall

scispacy/hearst_patterns.py
DeNeutoy: Ya, the approach here is to identify individual nouns, and then expand them in to noun chunks afterward, for exactly this reason.

scispacy/hyponym_detector.py
DeNeutoy: This is adding the token level attributes to only the noun compound's head, but then below use the noun chunk for the document level attribute. I think this is reasonably consistent with spacy's other methods, and also it protects against a token within a noun phrase having a hyponym which is not correct, e.g you could imagine `a country such as Great Britain`, would then get `Great._.hypernym == "country"` which isn't necessarily helpful.

danielkingai2: hmm, `Britain._.hypernym == "country"` isn't particularly helpful either though, since the full phrase is `Great Britain`. I can't imagine ever wanting to use hypernym/hyponym without getting the whole phrase. Can you? My soft preference would be for hypernym/hyponym attributes to always point to the full noun phrase. The way its currently set up, if I iterate over the doc and check if each token has a hypernym, i can't really get the whole phrase without replicating the noun phrase expansion logic, right? I think it would be better to be able to get the full phrase when iterating over the doc one token at a time.

DeNeutoy: Maybe we should just remove the token level attributes entirely, as I don't think that behaviour is ever desirable/that useful? It's kind of a span level attribute anyway?

danielkingai2: I can imagine wanting to a know if a specific token participates in a h*nym. If we drop the token level attribute completely, to determine this, you would need to create a `set` out of the token indices (or `dict`) that participate in h*nyms and then check against that `set` (or `dict`). If forcing people to do this themselves sounds fine to you, I'm good with just dropping the token level attributes, but I could imagine having this kind of easy lookup from a token be useful. For example, the lack of this is exactly why writing a spacy matcher for hearst patterns is hard, because tokens don't say if they participate in a noun phrase. Separately, I realized we could actually do this with a matcher pretty easily if we preprocessed to add a token attribute that said whether a token is part of a noun phrase or not, right?

danielkingai2: EDIT: nvm, you can't match on custom attributes

DeNeutoy: Ok, I think i'll remove them, because it would be weird to have token level attributes for span based properties - spacy doesn't do this (e.g you can't access an NER span from a token). Also it's relatively easy to do this afterward, via span indices.

PR #280, opened 2020-10-16
Release v0.3.0

DeNeutoy:
Updates for the release - i'm going to bump to `0.3.0-unreleased` in a separate PR, just in case the new github actions thing to publish to pypi goes wrong and I need to fix it, but normally we would be able to just roll that up into the same PR and choose the commit to tag as the release.

If you could try installing one of the models (I did and it was fine, but just to be sure) that would be great.


danielkingai2 (approved):
LGTM

PR #282, opened 2020-10-25
Added index checking to hyponym_detector.py

dumbshow:
Hi, I was getting errors with noun phrases that occurred at the beginning and end of strings in hyponym_detector.py.  (Not sure if I have filed this request correct....I am new to git....)  Thank you.

### DIFF
--- a/scispacy/hyponym_detector.py
+++ b/scispacy/hyponym_detector.py
@@ -62,6 +62,8 @@ class HyponymDetector:
 
         start = token.i
         while True:
+            if start==0:
+                break
             previous = doc[start - 1]
             if previous.pos_ in {"PROPN", "NOUN", "PRON"}:
                 start -= 1
@@ -70,7 +72,10 @@ class HyponymDetector:
 
         end = token.i + 1
         while True:
-            previous = doc[end]
+            try:
+                previous = doc[end]
+            except IndexError:
+                break
             if previous.pos_ in {"PROPN", "NOUN", "PRON"}:
                 end += 1
             else:


danielkingai2:
@DeNeutoy can you have a look a this?

scispacy/hyponym_detector.py
```diff
@@ -62,6 +62,8 @@ def expand_to_noun_compound(self, token: Token, doc: Doc):
 
         start = token.i
         while True:
+            if start==0:
```
DeNeutoy, line 65: Please add spaces between symbols, (our linter requires it), like `if start == 0:`

scispacy/hyponym_detector.py
```diff
@@ -70,7 +72,10 @@ def expand_to_noun_compound(self, token: Token, doc: Doc):
 
         end = token.i + 1
         while True:
-            previous = doc[end]
+            try:
```
DeNeutoy, line 75: Instead of using `try/except` here, could you get the length of the doc before this while loop (`len(doc)`) and check against it each time? It's more explicit and doesn't silently catch other index errors that may be thrown e.g within spacy.

DeNeutoy (commented):
@dumbshow, thanks for catching this! Two small things, and then also could you add a test for this in this file?

https://github.com/allenai/scispacy/blob/master/tests/test_hyponym_detector.py#L8

Thanks!

scispacy/hyponym_detector.py
dumbshow: Sorry for the delay.  Should have time to make these updates shortly.

DeNeutoy: No problem, contributions are always welcome regardless of timeframe :)

Issue #285, opened 2020-11-03
RAM Requirements


fschlatt:
When I try to load any SciSpaCy model with the EntityLinker, I am running out of memory. My current setup has 16 Gb. Since the 5 cached files for SciSpaCy are all below 1Gb I would expect it to run fine. Am I overlooking something? How much RAM is required to run the EntityLinker?

danielkingai2:
For me, loading the small model and the EntityLinker takes ~11gb of RAM. I don't have an immediate solution to make it use less, but we do also 4 more specific entity linkers (see the EntityLinker section of the readme) that are smaller than the large UMLS one. Not sure if @DeNeutoy has other ideas since he has been messing around with these more than me lately.

fschlatt:
What if all the lookup tables, like the `ann_concept_aliases_list` and the jsons in the knowledge are replaced by sqlite or other light-weight DBs. Keeping them in memory would of course be faster, but saving a lot of RAM for a slight performance decrease may make the EntityLinker more accessible for many people. Since I currently am unable to access a more powerful machine, I may try to implement the DB solution in my own fork. If and when I finish, I can give an update about how it works

danielkingai2:
Try again off of master? I reduced memory by a couple gigs in #287

danielkingai2:
btw, if this isn't enough and you want to mess around with reducing memory further, https://github.com/pythonprofilers/memory_profiler is a nice tool (and is how i made the graphs in that PR)

kaushikacharya:
@fschlatt
  Alternative approach could be building a custom UMLS subset in case there are [semantic types](https://metamap.nlm.nih.gov/SemanticTypesAndGroups.shtml) which you don't need.
Have a look at the thread: https://github.com/allenai/scispacy/issues/237

PR #287, opened 2020-11-10
Change load to generator

danielkingai2:
line 65 is the only actual change

Before:
out

After:
out2

Possibly a small change that allows entity linker to run on 16gb machine: #285


DeNeutoy (approved):
Nice, LGTM

Issue #242, opened 2020-06-24
UmlsEntityLinker file in cache


giuliacassara:
Hi! Everytime I run this line of code
linker = UmlsEntityLinker(resolve_abbreviations=True)
it will download the file tfidf_vectors_sparse.npz in a temporary folder. I want to avoid to download it every time, but I don't understand where the script is looking for the file. I was looking at your code https://github.com/allenai/scispacy/blob/master/scispacy/file_cache.py and I find this reference : SCISPACY_CACHE but I can't find it in my home path.

giuliacassara:
Ok, it is looking on ('~.scispacy'). I have downloaded the file tfidf_vectors_sparse.npz there but still, when I run linker = UmlsEntityLinker(resolve_abbreviations=True) it doesn't find anything in cache.

DeNeutoy:
@giuliacassara hmm, that is surprising. Can you paste the output of running it twice along with the contents of `ls ~/.scispacy` please?

GanjinZero:
The main reason of this error is scispacy check cache_path with a sha256 prefix in **file_cache.py**, putting **tfidf_vectors_sparse.npz** in **~/.scispacy/datasets** is not enough.

```python
filename = url_to_filename(url, etag)
cache_path = os.path.join(cache_dir, filename)
```

PR #288, opened 2020-11-23
allow file_cache detect local files correctly

GanjinZero:
When one download files for EntityLinker and put it into `~\.scispacy\datasets\`. Scispacy will check the cached files with a sha256 prefix.
In this pull request scispacy will check if `tfidf_vectors_sparse.npz` or `sha256 + tfidf_vectors_sparse.npz` exists under  `~\.scispacy\datasets\`.
Solve problem for #242

### DIFF
--- a/scispacy/file_cache.py
+++ b/scispacy/file_cache.py
@@ -123,7 +123,9 @@ def get_from_cache(url: str, cache_dir: str = None) -> str:
     # get cache path to put the file
     cache_path = os.path.join(cache_dir, filename)
 
-    if not os.path.exists(cache_path):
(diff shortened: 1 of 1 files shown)


DeNeutoy (commented):
I understand the motivation behind this change, but really this isn't the right way to go about this. We want the file cache to be for cached files downloaded by `cached_path` only, as otherwise it's kind of broken.

Instead, you could add another entry here: 

https://github.com/allenai/scispacy/blob/master/scispacy/candidate_generation.py#L79

to locally downloaded files. Would this address your need to use cached files?

PR #293, opened 2020-12-25
Windows utf8

### LINKED ISSUE #292: Error while running test_data_util.py
Function: test_example_iterator()
Assertion fails at assert entity.mention_text == example.text[entity.start: entity.end] because
entity.mention_text = polished silicon
example.text[entity.start: entity.end] = d polished silic
Meaning, indices entity.start: entity.end are wrong?

danielkingai2:
I know we don't really support windows, just fixing this one since someone opened an issue and it was easy.

Fixes #292

### DIFF
--- a/scispacy/data_util.py
+++ b/scispacy/data_util.py
@@ -44,7 +44,6 @@ def process_example(lines: List[str]) -> MedMentionExample:
         entities.append(
             MedMentionEntity(int(start), int(end), mention, mention_type, umls_id)
         )
-
     return MedMentionExample(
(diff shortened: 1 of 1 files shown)


danielkingai2:
I have no idea why the CI is failing. It can't find numpy.

yogeshhk:
The encoding change has fixed the failure on windows. Checked locally. No test is failing now.

DeNeutoy (approved):
Sorry for the delay, LGTM

PR #294, opened 2020-12-27
update dockerfile python and debian

danielkingai2:
CI is failing because the python/debian combination in the Dockerfile doesn't have a wheel for the latest spacy, so its trying to build it or something. Seems reasonable to update to python 3.8 and latest stable debian anyway.

Should unblock #293


DeNeutoy (approved):
LGTM

PR #295, opened 2020-12-30
upgrading to spacy 3

### LINKED ISSUE #303: Support spaCy 3.0
spaCy was upgraded to 3.0 this week and it appears the current version of scispacy is incompatible. I see there is a spacy3 branch and #295 appears to be a WIP on the upgrade. 

Looks like a big pull on the move to 3.0. Do you have an idea on the timeline for the upgrade? Thank you for putting this project out there!

Here is the stack using a Python 3.8 virtualenv with spacy 3.0.1 and en_core_sci_md 0.3.0 installed. 

```
>>> import en_core_sci_md
/tmp/test/lib/python3.8/site-packages/spacy/util.py:715: UserWarning: [W094] Model 'en_core_sci_md' (0.3.0) 
specifies an under-constrained spaCy version requirement: >=2.3.1. This can lead to compatibility problems with 
older versions, or as new spaCy versions are released, because the model may say it's compatible when it's not. 
Consider changing the "spacy_version" in your meta.json to a version range, with a lower and upper pin. 
For example: >=3.0.1,<3.1.0
  warnings.warn(warn_msg)
>>> en_core_sci_md.load()
Traceback (most recent call last):
  File "<stdin>", line 1, in <module>
  File "/tmp/test/lib/python3.8/site-packages/en_core_sci_md/__init__.py", line 10, in load
    nlp = load_model_from_init_py(__file__, **overrides)
  File "/tmp/test/lib/python3.8/site-packages/spacy/util.py", line 514, in load_model_from_init_py
    return load_model_from_path(
  File "/tmp/test/lib/python3.8/site-packages/spacy/util.py", line 388, in load_model_from_path
    config = load_config(config_path, overrides=dict_to_dot(config))
  File "/tmp/test/lib/python3.8/site-packages/spacy/util.py", line 545, in load_config
    raise IOError(Errors.E053.format(path=config_path, name="config.cfg"))
OSError: [E053] Could not read config.cfg from /tmp/test/lib/python3.8/site-packages/en_core_sci_md/
en_core_sci_md-0.3.0/config.cfg
```

danielkingai2:
- [x] add lemmatizer and attribute_ruler to configs
- [x] retrain
- [x] update docs to not add those pipes
- [x] update performance numbers in docs
- [x] retest the hyponym detector and make sure it still works
- [x] reupload models
- [ ] transformer model (separate pr)

Closes #303


project.yml
```diff
@@ -0,0 +1,354 @@
...
+    - evaluate-ner-sm
+    - evaluate-ner-md
+    - evaluate-ner-lg
+    - evaluate-specialized-ner
+
+commands:
+  - name: download
+    help: "Download the necessary files"
+    script:
+      - "aws s3 cp s3://ai2-s2-scispacy/data/gorc_subset.freqs assets/gorc_subset.freqs --no-sign-request"
```
DeNeutoy, line 69: Is there a reason these can't be top level `assets` which would be fetched by running `spacy assets`? https://nightly.spacy.io/usage/projects#assets

Is it because of the tar file?

project.yml
```diff
@@ -0,0 +1,354 @@
...
+    outputs:
+      - "project_data/genia_train.spacy"
+      - "project_data/genia_dev.spacy"
+      - "project_data/genia_test.spacy"
+      - "project_data/train"
+      - "project_data/dev"
+      - "project_data/test"
+
+  - name: convert-sm
+    help: "Convert the data to spaCy's format"
```
DeNeutoy, line 125: How about - "convert a file of word/document frequency statistics to spacy's vocabulary format."

project.yml
```diff
@@ -0,0 +1,354 @@
...
+vars:
+  freqs_loc: "https://s3-us-west-2.amazonaws.com/ai2-s2-scispacy/data/gorc_subset.freqs"
+  vectors_loc: "https://s3-us-west-2.amazonaws.com/ai2-s2-scispacy/data/pubmed_with_header.txt.gz"
+
+# These are the directories that the project needs. The project CLI will make
+# sure that they always exist.
+directories: ["assets", "project_data", "output"]
+
+workflows:
+  all-sm:
```
DeNeutoy, line 13: These could be called small, medium and large rather than `all-*`

configs/ner_sm.cfg
```diff
@@ -0,0 +1,137 @@
...
+
+[components.ner.model.tok2vec.encode]
+@architectures = "spacy.MaxoutWindowEncoder.v1"
+width = 256
+depth = 8
+window_size = 1
+maxout_pieces = 3
+
+[components.parser]
+source = "output/en_core_sci_sm_parser_tagger/model-best"
```
DeNeutoy, line 54: oof, this is pretty hard to tie back to the project.yml file. Is there a way to pass these components in via the `spacy train` cli instead?

scispacy/base_project_code.py
```diff
@@ -0,0 +1,109 @@
+from typing import Optional, Callable
+from pathlib import Path
```
DeNeutoy, line 2: This file is missing type hints in some places, but otherwise I like how straightforward adding in our custom readers etc is. This was my main concern about how we would use the projects stuff. Nice one for figuring it out!

scripts/convert_freqs.py
```diff
@@ -0,0 +1,79 @@
+import argparse
```
DeNeutoy, line 1: This script is getting quite a few steps removed from the original corpus now. It might be better to convert this to a script which reads a large corpus and creates the vocabularies directly, rather than us having created this intermediate file with the word/doc counts in it, and then having this one generate a vocabulary file which is not substantially different apart from how it is filtered.

scripts/evaluate_ner.py
```diff
@@ -0,0 +1,43 @@
...
+
+import spacy
+import importlib
+
+from scispacy.data_util import read_full_med_mentions, read_ner_from_tsv
+from scispacy.train_utils import evaluate_ner
+
+
+def main(model_path: str, dataset: str, output_path: str, code: Optional[str]):
+    if code is not None:
```
DeNeutoy, line 11: Does this happen automatically inside the created models? Presumably, right? Is it just because we haven't packaged them at this point that we need this additional code injection?

configs/ner_bc5cdr.cfg
```diff
@@ -0,0 +1,135 @@
+[paths]
```
DeNeutoy, line 1: Having all these config files for the ner models is not super great, as presumably they are identical apart from the paths. Is there a way to generate these templates programmatically somehow so we can avoid this?

DeNeutoy (commented):
Mostly questions so far, but looking great!

project.yml
danielkingai2: it is because i didn't notice that command lol, it might just work

configs/ner_sm.cfg
danielkingai2: Yeah, looks like overrides can be passed in https://nightly.spacy.io/api/cli#train. seems better than this, i agree

scripts/convert_freqs.py
danielkingai2: I never really read this code, just copied it around, so you're probably correct

scripts/evaluate_ner.py
danielkingai2: Right, this is exactly what all the spacy CLI commands that accept `--code` do. Also we're actually going to be able to get ride of the `--code` bit (except for training) after rc3 is released, because the tokenizer callback will be in a different place such that it only needs to happen when our base model is made. And, I wanted to eval the models before packaging separately from after packaging because of that one time we had a bug that caused a difference

configs/ner_bc5cdr.cfg
danielkingai2: I think we can use https://nightly.spacy.io/api/cli#init-fill-config to eliminate the boiler plate repetition, just haven't looked into it yet.

danielkingai2: used overrides and reduced to three distinct config files

configs/base_ner.cfg
```diff
@@ -0,0 +1,138 @@
...
+ents_f = 1.0
+ents_p = 0.0
+ents_r = 0.0
+
+[pretraining]
+
+[initialize]
+vectors = ${paths.vectors}
+init_tok2vec = ${paths.init_tok2vec}
+vocab_data = null
```
danielkingai2, line 133: Note to self: should this be a path to the vocab file?

configs/base_parser_tagger.cfg
```diff
@@ -0,0 +1,162 @@
+[paths]
+genia_train = "project_data/genia_train.spacy"
+genia_dev = "project_data/genia_dev.spacy"
+onto_train = "project_data/train"
```
danielkingai2, line 4: Note to self: should i pass these in as overrides?

configs/base_parser_tagger.cfg
```diff
@@ -0,0 +1,162 @@
...
+ents_f = 0.0
+ents_p = 0.0
+ents_r = 0.0
+
+[pretraining]
+
+[initialize]
+vectors = ${paths.vectors}
+init_tok2vec = ${paths.init_tok2vec}
+vocab_data = null
```
danielkingai2, line 157: Note to self: should this be a path to the vocab file?

configs/base_specialized_ner.cfg
```diff
@@ -0,0 +1,138 @@
...
+ents_f = 1.0
+ents_p = 0.0
+ents_r = 0.0
+
+[pretraining]
+
+[initialize]
+vectors = ${paths.vectors}
+init_tok2vec = ${paths.init_tok2vec}
+vocab_data = null
```
danielkingai2, line 133: Note to self: should this be a path to the vocab file?

configs/base_parser_tagger.cfg
danielkingai2: eh, not doing

configs/base_parser_tagger.cfg
danielkingai2: done

configs/base_specialized_ner.cfg
danielkingai2: done

configs/base_ner.cfg
danielkingai2: done

scripts/convert_freqs.py
danielkingai2: #300

project.yml
danielkingai2: i'm going to leave this as is, since i don't see any particular advantage to using the `assets` command. feel free to unresolve if you disagree.

DeNeutoy: Sure, it was a nit for sure

scispacy/abbreviation.py
```diff
@@ -145,14 +147,12 @@ class AbbreviationDetector:
     Note that this class does not replace the spans, or merge them.
     """
 
-    def __init__(self, nlp) -> None:
+    def __init__(self, nlp: Language, name: str = "") -> None:
```
DeNeutoy, line 150: Why the blank string rather than `abbreviation_detector`?

scispacy/linking.py
```diff
@@ -62,20 +64,24 @@ class EntityLinker:
 
     def __init__(
         self,
+        nlp: Language = None,  # required arg for spacy
+        name: str = "",  # required arg for spacy
```
DeNeutoy, line 68: same here, can these be their actual name, or are they overridden or something?

README.md
```diff
@@ -204,7 +213,7 @@ print("Name: ", entity)
 # Each entity is linked to UMLS with a score
 # (currently just char-3gram matching).
 for umls_ent in entity._.kb_ents:
-	print(linker.kb.cui_to_entity[umls_ent[0]])
+	print(nlp.get_pipe("scispacy_linker").kb.cui_to_entity[umls_ent[0]])
```
DeNeutoy, line 216: Nit: can you pull this `get_pipe` out of the loop, it makes it clearer to read

scispacy/hyponym_detector.py
```diff
@@ -22,11 +22,15 @@ class HyponymDetector:
 
     The pipe can be used with an instantiated spacy model like so:
     ```
-    hyponym_pipe = HyponymDetector(extended=True)
-    nlp.add_pipe(hyponym_pipe, last=True)
+    # you first need to add the lemmatizer from the base model
+    nlp_sm = spacy.load('en_core_web_sm')
+    nlp.add_pipe('attribute_ruler', source=nlp_sm, after='tagger')
```
DeNeutoy, line 27: Is the `attribute_ruler` actually required here, because we create our own `Matcher` instance below? If it is required, we should probably use the component that it creates, rather than create a new one here. (to be clear I think the Matcher inside the init here is fine)

DeNeutoy (commented):
Mostly looks good, some questions about adding the extra components

scispacy/abbreviation.py
danielkingai2: it can be whatever we want. the arg only gets used when spacy internals calls this `__init__` function, and it is set via the `Language.factory` at the top of the class. i only gave it a default here, because sometimes we initialize this class ourselves, and it is kind of weird to have to pass in a name that we never use. but the default can be anything.

scispacy/linking.py
danielkingai2: same as above, i can change them to be the actual names if you think that is nicer

README.md
danielkingai2: sure, will do

scispacy/hyponym_detector.py
danielkingai2: so, the `Matcher` requires a `lemmatizer`. In turn, the `lemmatizer` requires a `morphologizer`. the alternative to a `morphologizer` is a `tagger` followed by an `attribute_ruler`. the `attribute_ruler` has the rules for setting `pos` and `morph`, whereas the tagger just sets `tag`.

DeNeutoy: So if we just include the lemmatizer in the base model, this would all go away?

danielkingai2: yeah, would include the lemmatizer and the attribute_ruler. this would give us the rule based lemmatization again

DeNeutoy: gotcha, that makes me even more sure that figuring that out is worthwhile 👍

danielkingai2: It is either as easy as tweaking the config and rerunning the whole pipeline, or there is rake waiting for me to step on it, we shall see

scispacy/linking.py
DeNeutoy: Ya names plz 🙏

tests/test_hyponym_detector.py
```diff
@@ -6,29 +6,31 @@
...
 class TestHyponymDetector(unittest.TestCase):
-
     def setUp(self):
         super().setUp()
         self.nlp = spacy.load("en_core_sci_sm")
         self.detector = HyponymDetector(self.nlp, extended=True)
-        self.nlp.add_pipe(self.detector, last=True)
+
+        nlp_sm = spacy.load("en_core_web_sm")
+        self.nlp.add_pipe("attribute_ruler", source=nlp_sm, after="tagger")
```
DeNeutoy: We shouldn't need this anymore, right?

DeNeutoy (approved):
LGTM, assuming the tests pass without adding the lemmatizer!

tests/test_hyponym_detector.py
DeNeutoy: Lol you are too quick for me 👍

danielkingai2: Those emails from github "PR run failed" hurts too much

PR #298, opened 2021-01-07
Sort the candidates in entity linker.

wlwg:
The order of candidates returned from the UMLS candidate generator could be different between retries, so that when different candidates have the same scores, entity linker could return different results for each retry because of the `max_entities_per_mention` limit.

This PR is to sort the candidates from the candidate generator, so that the entity linker returns consistent results between retries.

### DIFF
--- a/scispacy/candidate_generation.py
+++ b/scispacy/candidate_generation.py
@@ -351,6 +351,7 @@ class CandidateGenerator:
                 MentionCandidate(concept, mentions, concept_to_similarities[concept])
                 for concept, mentions in concept_to_mentions.items()
             ]
+            mention_candidates = sorted(mention_candidates, key=lambda c: c.concept_id)
 
             batch_mention_candidates.append(mention_candidates)
 
--- a/tests/test_candidate_generation.py
+++ b/tests/test_candidate_generation.py
@@ -31,7 +31,7 @@ class TestCandidateGeneration(unittest.TestCase):
         results = candidate_generator(['(131)I-Macroaggregated Albumin'], 10)
 
         canonical_ids = [x.concept_id for x in results[0]]
-        assert canonical_ids == ['C0000005', 'C0000015', 'C0000102', 'C0000103', 'C0000074']
+        assert canonical_ids == ['C0000005', 'C0000015', 'C0000074', 'C0000102', 'C0000103']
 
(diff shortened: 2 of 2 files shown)


wlwg:
@danielkingai2 Hey Daniel. I'd appreciate it if you could quickly have a look at this PR.

danielkingai2:
@wlwg Could you share an example that produces different results on retries? I am not able to reproduce this locally, and tie-breaking in sorting is deterministic in python, so it would produce different results if nmslib is returning results in a different order each time (which it might be, but I'm not aware of this)

wlwg:
@danielkingai2 This is my code example. Run it multiple times can produce results with different order.

```
from scispacy.candidate_generation import CandidateGenerator

candidate_generator = CandidateGenerator(name='umls')
candidates = candidate_generator(['Benlysta'], 30)[0]
print(','.join([c.concept_id for c in candidates]))
```

danielkingai2:
@wlwg I still can't reproduce this, I ran 
```
In [17]: candidates = candidate_generator(['Benlysta'], 30)[0] 
    ...: for i in range(1000): 
    ...:     candidates_loop = candidate_generator(['Benlysta'], 30)[0] 
    ...:     assert candidates_loop == candidates 
```

without hitting the assert. Is it possible you are changing something else between runs? What version of scispacy and nmslib do you have?

wlwg:
@danielkingai2 You can't reproduce it by having a loop like that. (I assume it's because some kind of random seed is fixed during the whole program). You need to put it into a python script and run it multiple times in command line. You'll be able to see different results.

danielkingai2:
Ok, I was able to reproduce it, thanks. I believe it is caused by [this](https://github.com/allenai/scispacy/blob/7c852a4c34b09a68ca3db2e01dca09628ab1b425/scispacy/candidate_generation.py#L350-L353) iteration over a `defaultdict`. I think the candidate generator is the proper place to fix, rather than the linker. Would you be up for changing that loop to iterate in a deterministic order (probably just sort the dictionary keys), and verifying that fixes the issue? Thanks!

wlwg:
@danielkingai2 Thanks for confirming. I also agree that the change is better to be in candidate generator. I've made the changes and fixed a related unit test. Would appreciate it if you could review it.

danielkingai2 (approved):
LGTM, thanks!

PR #367, opened 2021-06-21
Add m parameter documentation

### LINKED ISSUE #366: Explanation for the M parameter in CreateIndex in nmslib
I was digging through code on creating the entity linker and found this TODO:

https://github.com/allenai/scispacy/blob/2290a80cfe0948e48d8ecfbd60064019d57a6874/scispacy/candidate_generation.py#L387

So i went on a deeper dive and found:

M param explanation

The M parameter seems to be the size of the neighbourhood in the underlying graph structure when creating the HNSW ( https://arxiv.org/ftp/arxiv/papers/1603/1603.09320.pdf) graph. 

My understanding is that putting this number higher increases the connectivity between clusters allowing for easier between-cluster jumping. 

The way they describe it in NMSLIB repo is:

"""
In what follows, we discuss HNSW-specific parameters. First, for HNSW, the parameter M defines the maximum number of neighbors in the zero and above-zero layers. However, the actual default maximum number of neighbors for the zero layer is 2*M. This behavior can be overriden by explicitly setting the maximum number of neighbors for the zero (maxM0) and above-zero layers (maxM). Note that the zero layer contains all the data points, but the number of points included in other layers is defined by the parameter mult (see the HNSW paper for details).
"""

MichalMalyska:
Closes #366

### DIFF
--- a/scispacy/candidate_generation.py
+++ b/scispacy/candidate_generation.py
@@ -388,7 +388,7 @@ def create_tfidf_ann_index(
     # by consecutive insertion of elements in a random order by connecting them to M closest neighbours
     # from the previously inserted elements. These later become bridges between the network hubs that 
     # improve overall graph connectivity. (bigger M -> higher recall, slower creation)
-    # For more details see: https://arxiv.org/ftp/arxiv/papers/1603/1603.09320.pdf
+    # For more details see:  https://arxiv.org/pdf/1603.09320.pdf?
     m_parameter = 100
     # `C` for Construction. Set to the maximum recommended value
     # Improves recall at the expense of longer indexing time
--- a/scispacy/candidate_generation.py
+++ b/scispacy/candidate_generation.py
@@ -384,7 +384,11 @@ def create_tfidf_ann_index(
     # Default values resulted in very low recall.
 
     # set to the maximum recommended value. Improves recall at the expense of longer indexing time.
-    # TODO: This variable name is so hot because I don't actually know what this parameter does.
+    # We use the HNSW (Hierarchical Navigable Small World Graph) representation which is constructed
+    # by consecutive insertion of elements in a random order by connecting them to M closest neighbours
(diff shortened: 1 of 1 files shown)


scispacy/candidate_generation.py
```diff
@@ -384,7 +384,11 @@ def create_tfidf_ann_index(
     # Default values resulted in very low recall.
 
     # set to the maximum recommended value. Improves recall at the expense of longer indexing time.
-    # TODO: This variable name is so hot because I don't actually know what this parameter does.
+    # We use the HNSW (Hierarchical Navigable Small World Graph) representation which is constructed
+    # by consecutive insertion of elements in a random order by connecting them to M closest neighbours
+    # from the previously inserted elements. These later become bridges between the network hubs that 
+    # improve overall graph connectivity. (bigger M -> higher recall, slower creation)
+    # For more details see: https://arxiv.org/ftp/arxiv/papers/1603/1603.09320.pdf
```
danielkingai2, line 391: Any reason for the long url instead of https://arxiv.org/pdf/1603.09320.pdf?

danielkingai2 (approved):
Thanks!

scispacy/candidate_generation.py
MichalMalyska: Nope, I just copypasted from my issue. Fixed now

MichalMalyska:
I think it will fail on linting again, do you have a script to run linting or should I just manually do it ?

danielkingai2:
The CI workflow is here (https://github.com/allenai/scispacy/blob/main/.github/workflows/main.yml). We use `black` for autoformatting, but the line length of comments isn't formatted by `black`, so fixing that is manual.

MichalMalyska:
@danielkingai2 fixed now, I also addressed a bunch of these in #368

MichalMalyska:
@danielkingai2 this should be ready now.

Issue #338, opened 2021-03-17
Span is not serializable in abbreviations - figure out a better workaround


f0lie:
```python
import spacy

from scispacy.abbreviation import AbbreviationDetector

nlp = spacy.load("en_core_sci_sm")

# Add the abbreviation pipe to the spacy pipeline.
nlp.add_pipe("abbreviation_detector")

test = ["Spinal and bulbar muscular atrophy (SBMA) is an inherited motor neuron disease caused by the expansion of a polyglutamine tract within the androgen receptor (AR). SBMA can be caused by this easily."]

print("Abbreviation", "\t", "Definition")
for doc in nlp.pipe(test, n_process=4):
    for abrv in doc._.abbreviations:
        print(f"{abrv} \t ({abrv.start}, {abrv.end}) {abrv._.long_form}")
```

Running that code leads to this. The error message doesn't make a lot of sense, It could be because there are more processes than entries. If you remove `n_process` the solves the problem.

```
Abbreviation     Definition
Abbreviation     Definition
Traceback (most recent call last):
  File "<string>", line 1, in <module>
  File "C:\Python38\lib\multiprocessing\spawn.py", line 116, in spawn_main
    exitcode = _main(fd, parent_sentinel)
  File "C:\Python38\lib\multiprocessing\spawn.py", line 125, in _main
    prepare(preparation_data)
  File "C:\Python38\lib\multiprocessing\spawn.py", line 236, in prepare
    _fixup_main_from_path(data['init_main_from_path'])
  File "C:\Python38\lib\multiprocessing\spawn.py", line 287, in _fixup_main_from_path
    main_content = runpy.run_path(main_path,
  File "C:\Python38\lib\runpy.py", line 265, in run_path
    return _run_module_code(code, init_globals, run_name,
  File "C:\Python38\lib\runpy.py", line 97, in _run_module_code
    _run_code(code, mod_globals, init_globals,
  File "C:\Python38\lib\runpy.py", line 87, in _run_code
    exec(code, run_globals)
  File "C:\Users\alexd\Dropbox (UFL)\UFII_COVID19_RESEARCH_TOPICS\cord19\text_parsing_pipeline\test.py", line 13, in <module>
    for doc in nlp.pipe(test, n_process=4):
  File "C:\Python38\lib\site-packages\spacy\language.py", line 1475, in pipe
    for doc in docs:
  File "C:\Python38\lib\site-packages\spacy\language.py", line 1511, in _multiprocessing_pipe
[...]

PR #368, opened 2021-06-21
Add abbreviation serializations

MichalMalyska:
Solves #338

### DIFF
--- a/scispacy/abbreviation.py
+++ b/scispacy/abbreviation.py
@@ -153,12 +153,13 @@ class AbbreviationDetector:
     name: `str`, a required argument for spacy to use this as a factory
     """
 
-    def __init__(self, nlp: Language, name: str = "abbreviation_detector") -> None:
+    def __init__(self, nlp: Language, name: str = "abbreviation_detector", make_serializable: bool = False) -> None:
         Doc.set_extension("abbreviations", default=[], force=True)
         Span.set_extension("long_form", default=None, force=True)
 
         self.matcher = Matcher(nlp.vocab)
         self.matcher.add("parenthesis", [[{"ORTH": "("}, {"OP": "+"}, {"ORTH": ")"}]])
+        self.make_serializable = make_serializable
         self.global_matcher = Matcher(nlp.vocab)
 
     def find(self, span: Span, doc: Doc) -> Tuple[Span, Set[Span]]:
@@ -186,6 +187,9 @@ class AbbreviationDetector:
             for short in short_forms:
                 short._.long_form = long_form
                 doc._.abbreviations.append(short)
+        if self.make_serializable:
+            abbreviations = doc._.abbreviations
+            doc._.abbreviations = [self.make_short_form_serializable(abbreviation) for abbreviation in abbreviations]
         return doc
 
     def find_matches_for(
@@ -223,3 +227,20 @@ class AbbreviationDetector:
             self.global_matcher.remove(key)
 
         return list((k, v) for k, v in all_occurences.items())
+
+    def make_short_form_serializable(self, abbreviation):
+        """
+        Converts the abbreviations into a short form that is serializable to enable multiprocessing
+
+        Parameters
+        ----------
+        short : Span
+            Short for abbreviation
+        long_form : Span
+            Long form of the abbreviation
+        """        
+        long_form = abbreviation._.long_form
+        abbreviation._.long_form = None
+        serializable_abbr = {"short_text": abbreviation.text, "short_start": abbreviation.start, "short_end": abbreviation.end,
+                             "long_text": long_form.text, "long_start": long_form.start, "long_end": long_form.end}
+        return serializable_abbr
--- a/scispacy/abbreviation.py
+++ b/scispacy/abbreviation.py
@@ -153,7 +153,12 @@ class AbbreviationDetector:
     name: `str`, a required argument for spacy to use this as a factory
     """
 
-    def __init__(self, nlp: Language, name: str = "abbreviation_detector", make_serializable: bool = False) -> None:
+    def __init__(
+        self,
+        nlp: Language,
+        name: str = "abbreviation_detector",
+        make_serializable: bool = False,
+    ) -> None:
         Doc.set_extension("abbreviations", default=[], force=True)
         Span.set_extension("long_form", default=None, force=True)
 
@@ -189,7 +194,10 @@ class AbbreviationDetector:
                 doc._.abbreviations.append(short)
         if self.make_serializable:
             abbreviations = doc._.abbreviations
-            doc._.abbreviations = [self.make_short_form_serializable(abbreviation) for abbreviation in abbreviations]
+            doc._.abbreviations = [
+                self.make_short_form_serializable(abbreviation)
+                for abbreviation in abbreviations
+            ]
         return doc
 
     def find_matches_for(
@@ -238,9 +246,15 @@ class AbbreviationDetector:
             Short for abbreviation
         long_form : Span
             Long form of the abbreviation
-        """        
+        """
         long_form = abbreviation._.long_form
         abbreviation._.long_form = None
-        serializable_abbr = {"short_text": abbreviation.text, "short_start": abbreviation.start, "short_end": abbreviation.end,
-                             "long_text": long_form.text, "long_start": long_form.start, "long_end": long_form.end}
+        serializable_abbr = {
+            "short_text": abbreviation.text,
+            "short_start": abbreviation.start,
+            "short_end": abbreviation.end,
+            "long_text": long_form.text,
+            "long_start": long_form.start,
+            "long_end": long_form.end,
+        }
         return serializable_abbr
--- a/scispacy/abbreviation.py
+++ b/scispacy/abbreviation.py
@@ -151,6 +151,8 @@ class AbbreviationDetector:
 
     nlp: `Language`, a required argument for spacy to use this as a factory
     name: `str`, a required argument for spacy to use this as a factory
+    make_serializable: `bool`, a required argument for whether we want to use the serializable
+    or non serializable version.
     """
 
     def __init__(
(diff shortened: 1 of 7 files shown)


scispacy/abbreviation.py
```diff
@@ -223,3 +227,20 @@ def find_matches_for(
...
+
+        Parameters
+        ----------
+        short : Span
+            Short for abbreviation
+        long_form : Span
+            Long form of the abbreviation
+        """        
+        long_form = abbreviation._.long_form
+        abbreviation._.long_form = None
```
MichalMalyska, line 243: I'm not 100% sure this is necessary (I just copied it from the issue)

MichalMalyska:
@danielkingai2 I fixed a bunch of issues that somehow were not picked up in previous PRs:
1. One of the test cases was using the old way of checking for dependency parsing
2. Reformatted all with black with line-length = 88 (like in your CI checks)
3. Fixed "," without following whitespace in a list to pass linting

MichalMalyska:
The only change that is outstanding there is switch from
`assert doc.is_parsed`
to
`assert doc.has_annotation("DEP")`

in one of the tests, this way it doesn't produce a warning

scispacy/abbreviation.py
MichalMalyska: This is actually needed (tests don't pass without it)

MichalMalyska:
While making sure that all works in the linking with the serializable abbreviation detector I noticed a TODO for adding short forms to entity linking so i just added it.

tests/custom_tests/test_all_model.py
```diff
@@ -36,6 +36,12 @@ def test_custom_segmentation(combined_all_model_fixture):
...
-    assert doc.is_parsed
+    assert doc.has_annotation("DEP")
     assert doc[0].dep_ == "ROOT"
     assert doc[0].tag_ == "NN"
+
+def test_full_pipe_serializable(combined_all_model_fixture):
+    text = "Induction of cytokine expression in leukocytes (CEIL) by binding of thrombin-stimulated platelets. BACKGROUND: Activated platelets tether and activate myeloid leukocytes."
+    doc = [doc for doc in combined_all_model_fixture.pipe([text, text], n_process = 2)][0]
+    # If we got here this means that both model is serializable and there is an abbreviation that would break if it wasn't
+    assert len(doc._.abbreviations) > 0
```
danielkingai2, line 47: could you also test that the abbreviations are as you expect? That is, the short form has a textual long form, and the first abbreviation is the expected dictionary.

scispacy/linking.py
```diff
@@ -96,20 +96,26 @@ def __init__(
         self.umls = self.kb
 
     def __call__(self, doc: Doc) -> Doc:
-        mentions = []
+        mention_strings = []
         if self.resolve_abbreviations and Doc.has_extension("abbreviations"):
-
+            # This used to create a mentions list of spans, just to create a list of mention strings later on
             for ent in doc.ents:
-                # TODO: This is possibly sub-optimal - we might
```
danielkingai2, line 103: I don't think this is quite this straightforward. The later part (line 115 in original) assumes that `mention_strings` is the same length as `doc.ents`. I guess we're not testing the combination of the abbreviation detector and the linker very well...

I would just do the Span vs str check for now. But also if you could fix the test so your proposed code fails it (or explain why I'm thinking about it wrong), that would be great.

danielkingai2 (changes requested):

scispacy/linking.py
```diff
@@ -96,20 +96,26 @@ def __init__(
...
+                    # Short form
+                    mention_strings.append(ent.text)
+                elif isinstance(ent._.long_form, str):
+                    # Long form
+                    mention_strings.append(ent._.long_form)
+                    # Short form
+                    mention_strings.append(ent.text)
                 else:
-                    mentions.append(ent)
+                    # Short form
```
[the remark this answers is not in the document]
MichalMalyska: Added short forms as well

scispacy/linking.py
MichalMalyska: I'll make this simpler tomorrow and then start working on the more complex version when I have some more time

scispacy/linking.py
```diff
@@ -96,20 +96,26 @@ def __init__(
         self.umls = self.kb
 
     def __call__(self, doc: Doc) -> Doc:
-        mentions = []
+        mention_strings = []
         if self.resolve_abbreviations and Doc.has_extension("abbreviations"):
-
+            # This used to create a mentions list of spans, just to create a list of mention strings later on
```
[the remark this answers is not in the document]
MichalMalyska: Done

scispacy/abbreviation.py
```diff
@@ -223,3 +235,26 @@ def find_matches_for(
             self.global_matcher.remove(key)
 
         return list((k, v) for k, v in all_occurences.items())
+
+    def make_short_form_serializable(self, abbreviation):
```
danielkingai2, line 239: type hint on `abbreviation` please

scispacy/abbreviation.py
```diff
@@ -223,3 +235,26 @@ def find_matches_for(
...
+        """
+        Converts the abbreviations into a short form that is serializable to enable multiprocessing
+
+        Parameters
```
danielkingai2, line 243: adjust params to match the actual params

scispacy/linking.py
```diff
@@ -96,20 +96,21 @@ def __init__(
...
-        mentions = []
+        mention_strings = []
         if self.resolve_abbreviations and Doc.has_extension("abbreviations"):
-
             for ent in doc.ents:
-                # TODO: This is possibly sub-optimal - we might
-                # prefer to look up both the long and short forms.
-                if ent._.long_form is not None:
-                    mentions.append(ent._.long_form)
+                if isinstance(ent._.long_form, Span):
```
danielkingai2, line 102: could you add the TODO comment back? You can also remove the "Long form" comments below

tests/conftest.py
```diff
@@ -19,14 +20,15 @@ def get_spacy_model(
...
+    with_serializable_abbreviation_detector: bool = False,
 ) -> SpacyModelType:
     """
     In order to avoid loading spacy models repeatedly,
     we'll save references to them, keyed by the options
     we used to create the spacy model, so any particular
     configuration only gets loaded once.
     """
-    options = (spacy_model_name, pos_tags, parse, ner, with_custom_tokenizer, with_sentence_segmenter)
+    options = (spacy_model_name, pos_tags, parse, ner, with_custom_tokenizer, with_sentence_segmenter, with_serializable_abbreviation_detector)
```
danielkingai2, line 31: I don't think the `with_serializable_abbreviation_detector` param does exactly what it should. Setting it to `False` doesn't add the abbreviation detector at all, whereas I think setting it to False should add the non serializable version of the abbreviation detector. Perhaps two params `with_abbreviation_detector` and `serializable_abbreviation_detector`

tests/conftest.py
```diff
@@ -97,9 +101,13 @@ def test_model_dir():
 
 @pytest.fixture()
 def combined_all_model_fixture():
-    nlp = get_spacy_model("en_core_sci_sm", True, True, True, with_custom_tokenizer=True, with_sentence_segmenter=False)
+    nlp = get_spacy_model("en_core_sci_sm", True, True, True, with_custom_tokenizer=True, with_sentence_segmenter=False, with_serializable_abbreviation_detector=True)
     return nlp
 
+@pytest.fixture()
+def combined_all_model_fixture_old_abbrev():
```
danielkingai2, line 108: `combined_all_model_fixture_old_abbrev` -> `combined_all_model_fixture_non_serializable_abbrev`

tests/custom_tests/test_all_model.py
```diff
@@ -36,6 +37,25 @@ def test_custom_segmentation(combined_all_model_fixture):
...
+    abbrev = doc._.abbreviations[0]
+    assert abbrev["short_text"] == "CEIL"
+    assert abbrev["long_text"] == "cytokine expression in leukocytes"
+    assert doc[abbrev["short_start"] : abbrev["short_end"]].text == abbrev["short_text"]
+    assert doc[abbrev["long_start"] : abbrev["long_end"]].text == abbrev["long_text"]
+
+
+@pytest.mark.xfail
```
danielkingai2, line 56: I realized I actually asked for the wrong thing here, because I want to make sure this test fails, not just allow it to fail (which is what xfail does). What we actually want is `pytest.raises`. So:

```suggestion
def test_full_pipe_not_serializable(combined_all_model_fixture_old_abbrev):
    text = "Induction of cytokine expression in leukocytes (CEIL) by binding of thrombin-stimulated platelets. BACKGROUND: Activated platelets tether and activate myeloid leukocytes."
    # This line requires the pipeline to be serializable (because it uses 2 processes), so the test should fail here
    with pytest.raises(TypeError):
        doc = [doc for doc in combined_all_model_fixture_old_abbrev.pipe([text, text], n_process = 2)][0]
```

danielkingai2 (commented):
Functionally looks good now. Left a few minor comments, and a request to change the test set up. After those things, should be good to merge. Thanks for your help with this!

tests/custom_tests/test_all_model.py
MichalMalyska: This won't work - the TypeError will get raised inside of a subprocess so it won't get caught, and won't terminate.

danielkingai2: oh ok, can you just replace with whatever the actual error is?

MichalMalyska:
@danielkingai2 I just did a ton of digging - your proposed test will not succeed (it will just hang forever) on spacy <3.1.0 since that's when they added a fix for multiprocessing errors:
https://github.com/explosion/spaCy/releases/tag/v3.1.0
- in particular this PR https://github.com/explosion/spaCy/pull/8004

Which is incompatible with the current requirements file for scispacy. 
We can add it as a commented out option until the dependency is updated, and keep the xfail in the meantime?

danielkingai2:
Weird, ok, thanks. Yeah, you can just leave it as xfail, note in a comment that it should fail, and check and make sure it actually fails for this PR.

MichalMalyska:
Should I also add the test to add when upgrading to 3.1.0 ?

danielkingai2:
Sure

MichalMalyska:
The xfail test also hangs now (with the updated fixtures etc), but we know that it fails if I manually execute it inside of pdb
Screen Shot 2021-07-14 at 3 44 50 PM

MichalMalyska:
I figured out which of the internals will fail even without multiprocessing if the doc is not serializable and added that as a test instead.

danielkingai2 (approved):
LGTM, thanks!

PR #369, opened 2021-06-21
Test ci

danielkingai2:


danielkingai2:
@MichalMalyska after this passes CI and I merge this, CI should be passing on main, so you should only need to fix things related to your PRs.

MichalMalyska:
My black version was different which was causing the reformatting changes, fixed now! Thank you.

PR #411, opened 2022-01-11
Update spacy dependency version

MichalMalyska:
If we can get this done + release then scispacy=0.5.0 would be compatible with medspacy which it currently is not. Passes all tests so there shouldn't be any issues

### DIFF
--- a/setup.py
+++ b/setup.py
@@ -41,7 +41,7 @@ setup(
     packages=find_packages(exclude=["*.tests", "*.tests.*", "tests.*", "tests"]),
     license="Apache",
     install_requires=[
-        "spacy>=3.0.0,<3.1.0",
+        "spacy>=3.0.0,<3.2.0",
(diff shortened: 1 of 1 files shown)


dakinggg:
I'm not at the Allen institute anymore, so this might move pretty slowly. But the other thing that would need to be done is check whether the models need to be retrained. I guess to check this you would need to run the model evaluations in project.yml and make sure the metrics are the same for at least the small model or something, on spacy 3.1 and spacy 3.2

dakinggg:
I should also note that if you urgently need this for your workflow, you can make the change and build scispacy from source

qhreul:
@dakinggg Does that mean that the SciSpacy is no longer managed under the Allen Institute?

dakinggg:
It means the package is very lightly maintained right now.

dakinggg:
Closing because I am retraining the models for spacy 3.2

PR #439, opened 2022-07-06
encoding="utf-8"

jgroehm:
Thanks for the amazing library! 

Hit the following error using export_umls_json.py with UMLS 2022AA:
'charmap' codec can't decode byte 0x81 in position X: character maps to <undefined>

Resolved by adding the utf-8 encoding throughout umls_utils.

### DIFF
--- a/scispacy/candidate_generation.py
+++ b/scispacy/candidate_generation.py
@@ -16,7 +16,7 @@ from scispacy.linking_utils import (
     UmlsKnowledgeBase,
     Mesh,
     GeneOntology,
-    RxNorm,
+    RxNorm, 
     HumanPhenotypeOntology,
 )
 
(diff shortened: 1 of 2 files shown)


dakinggg (approved):
Thanks @jgroehm LGTM! The CI is failing because we format all files with `black`. If you reformat the file with `black candidate_generation.py --line-length 88`, it should be good to go.

dakinggg:
@jgroehm The black commit doesn't look right. It doesn't contain any reformatting, just a stray space added. Can you try again?
Command should be `black candidate_generation.py --line-length 88` and then add and commit that change.

jgroehm:
Sorry for the false start there, new to black, must not have set up the VSCode extension correctly. CLI appears to have done the trick. Thank you!

dakinggg:
Hey @jgroehm, sorry I forgot about this. Merging now, thanks!

PR #472, opened 2023-03-31
Update EntityLinker docstring

andyjessen:
This commit updates the docstring to include current supported linkers.


dakinggg (approved):
Thanks @andyjessen !

Issue #473, opened 2023-04-12
Remove NMSLIB dependency


nanthony007:
I'm not sure if this would be possible and what alternatives may even exist, BUT, due to years of inactivity and unresponsiveness on the primary nmslib maintainer's side (not faulting him), the nmslib dependency makes scispacy _very_ unaccessible to new users and, in fact, will remain completely inaccessible to users on new operating systems (Windows 11) or running modern versions of python (3.11).

Are there any possible alternatives for the few lines of code where this package uses nmslib?

From what I can see those are primarily two calls to `nmslib.init()` and otherwise type annotations.

Please advise, if possible I would love to help here but am not comfortable writing robust production C++ code nor am I an expert on the scispacy models themselves.

dakinggg:
Hi @nanthony007, replacing nmslib with another approximate nearest neighbor search library is certainly doable, but is a bit more involved than you might realize. The candidate generator (https://github.com/allenai/scispacy/blob/4f9ba0931d216ddfb9a8f01334d76cfb662738ae/scispacy/candidate_generation.py#L148) uses nmslib for the approximate nearest neighbor search, so we would need to swap that out for another library, which means recreating an index with a different library (https://github.com/allenai/scispacy/blob/4f9ba0931d216ddfb9a8f01334d76cfb662738ae/scispacy/candidate_generation.py#L365 for doing it with nmslib), then rewriting the code to load and use that index, and then evaluating the candidate generation to make sure speed and accuracy are still on par with the previous implementation. This is unfortunately not something I am likely to have time to do in the near future, but I will try.

That being said, I have recently installed nmslib successfully on Windows Subsystem for Linux with python 3.10. 3.11 likely does not work, as you say.

nanthony007:
I ideally wanted to include scispacy as a dependency of a package for more novice programmers to have some simple access to biomedical NER and using WSL and/or navigating dependency (python, scispacy, etc) versions seems like mental overhead I want to avoid.


Is there a way this model could be re-trained using spacy's new entity linker itself?  Could that accomplish the same NEL while benefiting from scispacy's models?

nanthony007:
I wonder if [annoy](https://github.com/spotify/annoy) could be a good fit for an alternative ANN index?

nanthony007:
Please see #481

nanthony007:
Closing due to no clear direction forward...

phaeta:
[...]

PR #481, opened 2023-05-10
Convert nmslib to annoy

nanthony007:
This PR removes the dependency on `nmslib` and switches to `annoy` by Spotify [link](https://github.com/spotify/annoy) which appears to be more actively maintained and is a C++ library. All of the changes happen inside `scispacy/candidate_generation.py` since, as far as I could tell, that is where `nmslib` was used.

**Important**: Due to `nmslib` not working on my personal computer and me not knowing how to evaluate these changes and the requirements (i.e. generating UMLS knowledge base, indices as are, new indices, and some evaluation metric) I have not actually run this code so it quite likely breaks. I wanted to move beyond an issue since I really like this project and hope to use it at UK and I think this PR serves as a proof of concept for how it may not be _that_ complex to switch off of `nmslib`... then again I could be completely wrong 😄 

A few important notes on `annoy` vs `nmslib`:

- annoy claims to support cosine distance but it was not an option, so it appears to be supported "via" euclidean distance
- annoy only supports two major configurable parameters
  - number of trees generates (`n_trees` at index build time)
  - `k_search` at search/run-time as configurable parameters
- annoy doesn't support batch adding/searching so loops were required

Finally, as a general note, other that the _specific_ `nmslib` usages I tried to leave the code as much unchanged as possible although it is now possible that certain checks/crashes may not occur.

Please advise on next steps, I am willing to test this on a different device (allowing `nmslib` install) but would need guidance on evaluation procedures/metrics.

I hope this PR succinctly shows that switching (at least) _may_ be possible without too much overhead and that would open up `scispacy` for usage on more platforms with more python versions.

### DIFF
--- a/scispacy/candidate_generation.py
+++ b/scispacy/candidate_generation.py
@@ -10,6 +10,8 @@ from sklearn.feature_extraction.text import TfidfVectorizer
 import nmslib
 from nmslib.dist import FloatIndex
 
+from annoy import AnnoyIndex
+
 from scispacy.file_cache import cached_path
 from scispacy.linking_utils import (
     KnowledgeBase,
@@ -265,14 +267,13 @@ class CandidateGenerator:
         # remove empty vectors before calling `ann_index.knnQueryBatch`
         vectors = vectors[empty_vectors_boolean_flags]
 
-        # call `knnQueryBatch` to get neighbors
-        original_neighbours = self.ann_index.knnQueryBatch(vectors, k=k)
 
-        neighbors, distances = zip(
-            *[(x[0].tolist(), x[1].tolist()) for x in original_neighbours]
-        )
-        neighbors = list(neighbors)
-        distances = list(distances)
+        # search_k is second adjustable parameter, but has sensible default
+        original_neighbours = [self.ann_index.get_nns_by_vector(v=v, n=k, include_distances=True) for v in vectors]
+
+        # zip function is not needed due to annoy producing slightly different data structure
+        neighbors = list(original_neighbours[0])
+        distances = list(original_neighbours[1])
 
         # neighbors need to be converted to an np.array of objects instead of ndarray of dimensions len(vectors)xk
         # Solution: add a row to `neighbors` with any length other than k. This way, calling np.array(neighbors)
@@ -454,14 +455,14 @@ def create_tfidf_ann_index(
 
     print(f"Fitting ann index on {len(concept_aliases)} aliases (takes 2 hours)")
     start_time = datetime.datetime.now()
-    ann_index = nmslib.init(
-        method="hnsw",
-        space="cosinesimil_sparse",
-        data_type=nmslib.DataType.SPARSE_VECTOR,
-    )
-    ann_index.addDataPointBatch(concept_alias_tfidfs)
-    ann_index.createIndex(index_params, print_progress=True)
-    ann_index.saveIndex(ann_index_path)
+    # docs say cosine supported (via euclidean?)
+    ann_index = AnnoyIndex(concept_alias_tfidfs.shape[1], "euclidean")
+    # cannot add batch so add one at a time 
+    for i, tfidf_vector in enumerate(concept_alias_tfidfs):
+        ann_index.add_item(i, tfidf_vector.toarray().flatten())
+    # this is adjustable and one of the only two major adjustable parameters 
+    ann_index.build(n_trees=10_000)
+    ann_index.save(ann_index_path)
     end_time = datetime.datetime.now()
     elapsed_time = end_time - start_time
     print(f"Fitting ann index took {elapsed_time.total_seconds()} seconds")


dakinggg:
Hi @nanthony007, thanks for making a start at this! If you're interested, I'm happy to help you through all the steps that would be necessary in order to fully replace nmslib with annoy, but there are a few :) Basically it comes down to rebuilding the indexes and then testing them. To rebuild the indexes, you would need to get the [create linker](https://github.com/allenai/scispacy/blob/main/scripts/create_linker.py) script to run and produce the necessary artifacts. The input paths for that script can be found [here](https://github.com/allenai/scispacy/blob/main/scispacy/linking_utils.py). Once you have the linkers created, we would need to test a few things.

1. The accuracy is roughly similar to the accuracy using nmslib. You can check that by running the [evaluation script](https://github.com/allenai/scispacy/blob/main/evaluation/evaluate_linker.py). The accuracy should be roughly as below for the main UMLS linker. Note: running the eval script (and development in general) will require that you install scispacy from source with `pip install -e .`.
```
Total entities:  70262
Correct at 1:  39170 Recall at 1:  0.5574848424468418
Correct at 2:  46132 Recall at 2:  0.6565711195240671
Correct at 10:  53370 Recall at 10:  0.759585551222567
Correct at 40:  55821 Recall at 40:  0.7944692721527995
```
2. The speed is roughly similar to nmslib. I haven't written a speed benchmarking script for the linker, but it should be fairly straightforward, and I can measure the speed of nmslib for the new method if needed.
3. The file size is roughly similar to nmslib. If annoy produces files that are much larger than the files that nmslib produces, that might be a problem. I don't have any reason to think it would, just something to check.

If we get through all of that, I can upload the new files to s3, and do a new release that removes the nmslib dependency.

dakinggg:
btw, none of those steps should require installing nmslib. You can edit the `setup.py` to remove the nmslib and add the annoy depdendency, and then install from source with `pip install -e .`. I can do any benchmarking against the nmslib version that is necessary.

nanthony007:
So upon some quick trials, the annoy index was taking ~5hours to create (by adding items) on my computer, which would crash when trying to build. Turns out it results in ~10GB file on disk when only building for 50k sample of tfidf-vectors.

I think [faiss](https://github.com/facebookresearch/faiss) may be another option that seems more similar in API at least to nmslib. Will try tomorrow.

dakinggg:
One issue you may run into is that `nmslib` has specific support for a sparse cosine similarity, while annoy says its works up to ~1000 dimensions, and faiss says it is for dense vectors.

nanthony007:
Okay, given that, I suppose we would expect both of these solutions to fall short. Is there a reason why we did not just use `scipy.sparse` and something from sklearn?  Not an expert on the implementation details so forgive me.

nanthony007:
If you don't see any immediate problems I could try KNeighbors from scikit learn? I know that is frequently used with tfidf vectors. Appears one of their metrics is cosine distance as well.

nanthony007:
Do you have an idea of how long evaluation script should take? That would give us a quick speed benchmark right? At least to know in the same general performance area....

dakinggg:
I would expect approximate nearest neighbors via nmslib to be substantially faster, and better for high dimensional tfidf data than the sklearn exact nearest neighbors algorithm. Running the evaluation script on my machine took ~3 minutes.

nanthony007:
You would be right... I ran the evaluation and it took almost 12 hours, with only a slight recall degradation. It did only take ~2minutes to create the model as opposed to 2 hours to create the index.

CleanShot 2023-05-16 at 07 11 50@2x

nanthony007:
If you don't have any recommendations I think you can close this PR and its corresponding issue... #473.  

I did have another question regarding replacing the current entity linker with spacy's built in knowledge base and entity linker that would thus remove the nmslib dependency as well but I was unsure of the viability of that at this time... 

Thanks for all your help on this.

nanthony007:
There seems to be no further suggestions so I'm going to close this and the linked issue #473

laulopezreal:
Dear both @nanthony007 @dakinggg,

I have been trying to remove the dependency from nmslib using scann by google. You can check the documentation [here](https://github.com/google-research/google-research/tree/master/scann).

I am managing to get similar recalls in the evaluation script. Is this something that you'd like to see in a PR, or what is the recommended way to start working towards a contribution?

I agree that this is a very useful library with great potential, but these dependency issues make its access quite difficult to the bioscience community of python users.

Looking forward to hearing your thoughts. Cheers!

nanthony007:
I'd probably start here with @dakinggg and post some of the evaluation script metrics (at least semi-formally) and timings (time was also a larger concern) before committing to a more formal PR with code etc etc. Thats my two cents though. 

Excited if you got this working!

laulopezreal:
hahaha I really liked "Thats my two cents though".

Thank you, your PR helped me a lot actually when I got started.  Do you know roughly how much the evaluation script should be taking?

nanthony007:
I think above it says about 3 minutes to run the whole script. That would just give you a ballpark estimate of the same general performance area.

laulopezreal:
3 minutes :O crazy! Thanks for pointing it out, I just read the comment above. Mine is taking hours.. I wonder what is that nmslib has that makes it so quick and how come libraries that are supposed to rank better in the benchmarks (like SCANN, pyNNdescent, etc.) take hours. 3 minutes against 8 hours is a huge difference. Do you guys have any intuition or ideas?

nanthony007:
I think it is something to due with the fact that these are _really_ sparse vectors, and that `nmslib` has some custom logic tailored to that use case. Although I could be wrong there.

laulopezreal:
@nanthony007 @dakinggg Hello! Sorry for the radio silence. I´ve been particularly busy at work but I finally have made some time to keep testing.
I have found out that, by using the BM25S implementation in https://github.com/xhluca/bm25s, the evaluation script runs in around 7 minutes with the next results:
´´´
Correct at 1:  28984 Recall at 1:  0.41251316501095897
Correct at 2:  42163 Recall at 2:  0.6000825481768239
Correct at 10:  44996 Recall at 10:  0.6404030628220091
Correct at 40:  44996 Recall at 40:  0.6404030628220091
´´´

There is plenty of room to tweak the parameters, so I expect an improvement in both, the running time and the recall metrics. Please let me know your thoughts, your feedback at this stage would be super useful.

In the meanwhile, I´ll start working on my fork to open a PR.

laulopezreal:
Update, this is the recall that I am getting at the moment:
```
Correct at 1:  36348 Recall at 1:  0.5173208846887365
Correct at 2:  42925 Recall at 2:  0.6109276707181691
Correct at 10:  50470 Recall at 10:  0.7183114628106231
Correct at 40:  52154 Recall at 40:  0.7422788989781105
```

Evaluation script takes around 7 minutes to run, but I think I can improve that a lot.

PR #511, opened 2024-03-19
Fix export_umls_json.py

ethanhkim:
Fixing the previous commit for export_umls_json.py to add non_suppressed into main func call.


dakinggg (approved):
Thank you!

PR #529, opened 2024-10-25
Add shorter version of pip installing nmslib from source

svlandeg:
Adding to the docs on alternatives to install `mslib`, I thought it would be useful to add the shorter variant
```
pip install "nmslib @ git+https://github.com/nmslib/nmslib.git/#subdirectory=python_bindings"
```
as taken from https://github.com/allenai/scispacy/issues/473#issuecomment-1590443024. Works for me on Windows 10 & Python 3.10.


dakinggg (approved):
Thanks Sofie!

PR #531, opened 2024-10-27
update readme to include installation method for nmslib on Apple M2 chip using python3.9

Archertakesitez:
updated README.md to include installation method of nmslib on Apple M2 chip, using Python3.9, and added an icon for my installation method.


dakinggg (approved):
Thank you!

PR #540, opened 2025-06-16
Code documentation for the scispacy project

ivanmilevtues:
This PR contains **diagrams for the scispacy codebase**. You can see how they **render in Github** here:
https://github.com/CodeBoarding/GeneratedOnBoardings/blob/main/scispacy/on_boarding.md

The scispacy codebase is loved by a lot of scientists. This said I am sure many of them want to add small modification to the codebase/understand it better. The goal of this PR is to help that process. We believe that **visuals give immedaite context on a new person**, and they can focus on the module (component) which interests them. I would love to hear what do you think about diagram first documentation for on-boarding and if it fits in your existing on-boarding processes.

Any feedback is more than welcome! We also have a **free github action** out now, which can automatically updated the docs (let me know what you think about that)

Full disclosure: we're trying to turn this into a startup, but we're still in a very early stage and figuring out what will actually be useful for people.

### DIFF
--- a/.codeboarding/Core Entity Linking System.md
+++ b/.codeboarding/Core Entity Linking System.md
@@ -1,0 +1,209 @@
+```mermaid
+
+graph LR
+
+    EntityLinker["EntityLinker"]
+
+    CandidateGenerator["CandidateGenerator"]
+
+    FileCache["FileCache"]
+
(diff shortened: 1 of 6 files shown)


laulopezreal (approved):
Very nice, I´ve always missed a diagram like this in the project. I wish I have seen it when I first started using scispacy.

PR #541, opened 2025-08-01
Modernize packaging

cthoyt:
This PR moves some setup configuration out of `setup.py` and into `pyproject.toml`


dakinggg (commented):
I don't think this PR is installable as is. I get ``pyproject.toml` is using the `[project]` table, but the required `project.version` field is neither set nor present in the `project.dynamic` list` from `uv pip install -e .`. I'd suggest this PR gets rid of setup.py entirely to avoid any issues between the two. Separately, CI doesn't catch this installation failure, so it'd be good to add a pip install command to CI to make sure that works.

setup.py
```diff
@@ -21,37 +20,7 @@
...
-    keywords=["bioinformatics nlp spacy SpaCy biomedical"],
-    classifiers=[
-        "Intended Audience :: Science/Research",
-        "Development Status :: 3 - Alpha",
-        "License :: OSI Approved :: Apache Software License",
-        "Programming Language :: Python :: 3.6",
-        "Topic :: Scientific/Engineering :: Artificial Intelligence",
-        "Topic :: Scientific/Engineering :: Bio-Informatics",
-    ],
     packages=find_packages(exclude=["*.tests", "*.tests.*", "tests.*", "tests"]),
```
dakinggg, line 24: i'm not sure whether this is still necessary or not (you can remove it, install with toml, and see if the test files are included or not). if it is, i believe the modern version is

```
[tool.setuptools.packages.find]
exclude = ["*.tests", "*.tests.*", "tests.*", "tests"]
```

setup.py
```diff
@@ -21,37 +20,7 @@
...
-        "requests>=2.0.0,<3.0.0",
-        "conllu",
-        "numpy",
-        "joblib",
-        "nmslib>=1.7.3.6; python_version < '3.11'",
-        "nmslib-metabrainz==2.1.3; python_version >= '3.11'",
-        "scikit-learn>=0.20.3",
-        "pysbd",
-    ],
     tests_require=["pytest", "pytest-cov", "flake8", "black", "mypy"],
```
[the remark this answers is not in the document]
cthoyt: Done in 608f23c

setup.py
cthoyt: Done in 9cb79a3

cthoyt:
> I don't think this PR is installable as is. I get ``pyproject.toml`is using the`[project]` table, but the required `project.version` field is neither set nor present in the `project.dynamic` list` from `uv pip install -e .`. I'd suggest this PR gets rid of setup.py entirely to avoid any issues between the two.

Hi @dakinggg, thanks for the prompt review! Sorry there were a few mistakes, I was flying blind without having CI running. I only just realized I could make a parallel PR within my own fork to get CI to run. Now, the existing CI is passing (though it required running black to get up to the new 2025 standard). You'll have to click the button to allow my PR to run on CI one more time, thank you!

> Separately, CI doesn't catch this installation failure, so it'd be good to add a pip install command to CI to make sure that works.

I would like to follow-up in in a different PR that extends the current testing and can address this.

dakinggg (approved):

PR #542, opened 2025-08-02
Document extending scispaCy with additional ontologies with PyOBO

cthoyt:
Following these improvements in scispaCy:

- https://github.com/allenai/scispacy/pull/544
- https://github.com/allenai/scispacy/pull/547
- https://github.com/allenai/scispacy/pull/549
- https://github.com/biopragmatics/pyobo/pull/437

This PR adds documentation on to the README to showcase how arbitrary ontologies and databases can be loaded for use as a scispaCy entity linker via PyOBO. It does this by:

1. Adding an end-to-end example for annotating genes using HGNC
2. Adds a link to more example usage in the PyOBO docs at https://pyobo.readthedocs.io/en/latest/scispacy.html

This also closes #331. `hgnc` can be replaced with `to` to use the Plant Trait Ontology, however I wasn't able to construct a convincing use case, so that can be left to the issue submitter.

PR #544, opened 2025-08-04
Enable creation of entity linker from a knowledge base

cthoyt:
This PR adds a convenience function for creating an `EntityLinker` instance from a `KnowledgeBase` instance that wraps the generation of TF-IDF vectorizations. It also adds documentation with the following end-to-end example:

```python
import spacy
from scispacy.linking import EntityLinker
from scispacy.linking_utils import UmlsKnowledgeBase

umls_kb = UmlsKnowledgeBase()
linker = EntityLinker.from_kb(umls_kb)
nlp = spacy.load("en_core_web_sm")
doc = linker(nlp(text))
```

This will be much more powerful if #542 is accepted, as it will allow for high-level creationg of entity linkers from arbitrary ontologies.


scispacy/linking.py
```diff
@@ -93,6 +110,26 @@ def __init__(
         self.filter_for_definitions = filter_for_definitions
         self.max_entities_per_mention = max_entities_per_mention
 
+    @classmethod
+    def from_kb(
+        cls,
+        kb: KnowledgeBase,
+        *,
+        directory: Optional[str] = None,
```
dakinggg, line 118: probably `ann_index_out_dir` is a better name?

dakinggg (approved):

scispacy/linking.py
```diff
@@ -1,9 +1,13 @@
+import tempfile
```
dakinggg, line 1: unused import?

dakinggg (approved):

PR #553, opened 2025-10-01
Update Python version in publish workflow

dakinggg:
- Updates python version in release action to 3.9, as the action failed with 3.7 (https://github.com/allenai/scispacy/actions/runs/18147037481/job/51650670084)
- Updates version to 0.6.1 so that I can publish a release without modifying the github action or deleting a tag.


kyleclo (approved):
lgtm

PR #556, opened 2025-11-16
Feature | Use numpy.float32 when scipy >= 1.11

Odessit007:


scispacy/candidate_generation.py
```diff
@@ -494,8 +489,9 @@ def create_tfidf_ann_index(
         )
         with open(concept_aliases_path, "w") as file:
             json.dump(concept_aliases, file)
+        on_disk_data_type = numpy.float32 if not scipy_supports_sparse_float16() else numpy.float16
```
dakinggg:
```suggestion
        on_disk_data_type = (
            numpy.float32 if not scipy_supports_sparse_float16() else numpy.float16
        )
```

PR #565, opened 2026-04-07
Fix numpy v2 incompatibility (#536)

### LINKED ISSUE #536: ValueError: numpy.dtype size changed, may indicate binary incompatibility. Expected 96 from C header, got 88 from PyObject
Thanks for this project. 

 Its proving hard for me to get everything working on my Windows 11 WLS2 Ubuntu 24.04.1 LTS with Python 3.10.16

I have scispacy installed.
I ran the following
`
import spacy
nlp = spacy.load("en_core_sci_md")

`
I got the Error: OSError: [E050] Can't find model 'en_core_sci_sm'. It doesn't seem to be a Python package or a valid path to a data directory.

I downloaded the model from the GitHub project homepage.

I then  used "uv add path_to_download"

I think this went correctly as I see :
```
Resolved 65 packages in 1.58s
      Built en-core-sci-md
```
When I run my code again I see 
`
ValueError: numpy.dtype size changed, may indicate binary incompatibility. Expected 96 from C header, got 88 from PyObject

`

Can you advise me on how to resolve this issue ?

Chessing234:
## Summary
- Removes the `numpy<2.0` upper bound constraint in `pyproject.toml` that was causing binary incompatibility errors (`numpy.dtype size changed`) with numpy v2
- The constraint was originally added while waiting for spacy and nmslib to support numpy v2; both now do (spacy >=3.7, nmslib >=2.1.2)
- Sets a minimum of `numpy>=1.24.0` to ensure a reasonably modern numpy version

Fixes #536

## Test plan
- [ ] Install scispacy with numpy v2 and verify no `numpy.dtype size changed` errors
- [ ] Run existing test suite to confirm no regressions

🤖 Generated with [Claude Code](https://claude.com/claude-code)


Powerkrieger:
Not sure if im completely off track with this, but: 

- i built scispacy locally with the removal of numpy constraint
- that allowed spacy 3.8
- when trying to run scispacy models of version 0.5.4 i had a "config type mismatch with spacy 3.8's stricter validation"
  - include_static_vectors = "True" should be a bool but is a string

Issue #534, opened 2024-11-22
UMLS Entity Linker throws BadZipFile error


markediger:
I am trying to run a basic example of the UMLS Entity Linker:

`import spacy
from scispacy.umls_linking import UmlsEntityLinker
nlp = spacy.load('en_core_sci_md')
linker = UmlsEntityLinker()

nlp.add_pipe(linker)
doc = nlp("Spinal and bulbar muscular atrophy (SBMA) is an \
           inherited motor neuron disease caused by the expansion \
           of a polyglutamine tract within the androgen receptor (AR). \
           SBMA can be caused by this easily.")

entity = doc.ents[1]
print("Name: ", entity)

for umls_ent in entity._.umls_ents:
    print(linker.umls.cui_to_entity[umls_ent[0]])`

I get an error implying that scispacy is not able to identify the UMLS dictionaries?

`Traceback (most recent call last):
  File "H:\integrated_evidence\indication_coding\indication-master\src\scispacy_test.py", line 5, in <module>
    linker = UmlsEntityLinker()
             ^^^^^^^^^^^^^^^^^^
  File "H:\integrated_evidence\indication_coding\indication-master\.venv\Lib\site-packages\scispacy\linking.py", line 85, in __init__
    self.candidate_generator = candidate_generator or CandidateGenerator(
                                                      ^^^^^^^^^^^^^^^^^^^
  File "H:\integrated_evidence\indication_coding\indication-master\.venv\Lib\site-packages\scispacy\candidate_generation.py", line 222, in __init__
    self.ann_index = ann_index or load_approximate_nearest_neighbours_index(
                                  ^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^
  File "H:\integrated_evidence\indication_coding\indication-master\.venv\Lib\site-packages\scispacy\candidate_generation.py", line 133, in load_approximate_nearest_neighbours_index
    concept_alias_tfidfs = scipy.sparse.load_npz(
                           ^^^^^^^^^^^^^^^^^^^^^^
  File "H:\integrated_evidence\indication_coding\indication-master\.venv\Lib\site-packages\scipy\sparse\_matrix_io.py", line 134, in load_npz
    with np.load(file, **PICKLE_KWARGS) as loaded:
         ^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^
[...]

PR #588, opened 2026-06-08
Raise informative IOError on corrupt tfidf vectors cache file

H2908:
Problem
When scipy.sparse.load_npz() encounters a corrupted .npz file (typically from an interrupted download), it raises BadZipFile, EOFError, or ValueError with no indication of which file is affected or how to fix it. Users are left with a cryptic traceback and no recovery path.
This was noted in issue #534 where a maintainer suggested modifying the package to print out which paths it is trying to load.
Fix
In load_approximate_nearest_neighbours_index:

Extract the resolved local cache path into a named variable before calling load_npz()
Catch BadZipFile, EOFError, and ValueError
Re-raise as IOError with the exact file path and instructions to delete the file so it can be re-downloaded

Testing
Added a test in tests/test_candidate_generation.py that writes a malformed .npz file to a temporary directory and asserts that the new IOError is raised.


H2908:
Hi, I'd like to work on this issue.
I've reproduced the problem — when scipy.sparse.load_npz() encounters a corrupted .npz file (typically from an interrupted download), it raises BadZipFile or EOFError with no indication of which file is affected or how to fix it.
My fix extracts the resolved local cache path into a variable before calling load_npz(), then catches BadZipFile, EOFError and ValueError, and re-raises as an IOError that includes the file path and instructs the user to delete it so it can be re-downloaded. I've also added a test that writes a malformed .npz file and asserts the new IOError is raised.
I'll open a PR shortly.

